PR #1, opened 2015-05-29
Fix Composer Dependency: EventManager is required for TableGateway

mwillbanks:
## Overview

As stated, this currently causes issues with anyone that is using TableGateway as the EventManager is a dependency.

## Workaround

```
composer require zendframework/zend-eventmanager
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,10 +14,10 @@
     },
     "require": {
         "php": ">=5.3.23",
+        "zendframework/zend-eventmanager": "~2.4",
         "zendframework/zend-stdlib": "~2.4"
     },
     "require-dev": {
-        "zendframework/zend-eventmanager": "~2.4",
         "zendframework/zend-mvc": "~2.4",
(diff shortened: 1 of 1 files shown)


mwillbanks:
ping @weierophinney @Maks3w

gianarb:
:+1: There are a lot of discussion about this type of issue..
http://gianarb.it/blog/the-price-of-modularity/ 

this PR is PERFECT!

weierophinney:
We already have a "suggest" entry for `zendframework/zend-eventmanager` (as well as `servicemanager`), and I'd argue that's the correct place to put it. However, I think we should rephrase the suggest entry to the following:

```json
"zendframework/zend-eventmanager": "@~2.4; required when using the TableGateway subcomponent"
```

and do similarly for the servicemanager.

Maks3w:
I'll close this for continue the discussion on #25.

PR #3, opened 2015-06-04
drop php 5.4 check and $that = $this usage

samsonasik:


src/Sql/Platform/Mysql/Ddl/CreateTableDecorator.php
```diff
@@ -163,14 +163,13 @@ private function normalizeColumnOption($name)
...
     /**
-     * @internal @private this method is only public for PHP 5.3 compatibility purposes.
      *
      * @param string $columnA
      * @param string $columnB
      *
      * @return int
      */
-    public function compareColumnOptions($columnA, $columnB)
+    private function compareColumnOptions($columnA, $columnB)
```
mwillbanks: I would instead make this protected, however, this is a BC break in the event someone extended the class.  You also cannot change visibility.  I would revert this portion of the commit so that it can be merged to master and subsequently make this change against the develop branch as a secondary PR.

Ocramius: The method had been marked as `@internal` and `@private` (I think by me). Can't have more warnings than that - `private` is perfectly fine here, and nobody should ever have relied on it in first place.

Ocramius: Yep, it was me: https://github.com/zendframework/zend-db/commit/bf6d9fc0ad72137739e7e51a3113bf6e32ce39cb

mwillbanks:
@Ocramius makes sense but still would need to be broken in a minor version vs. a bug fix release.

samsonasik:
should I change to protected, then?

Warm regards,

Abdul Malik Ikhsan

Pada 23 Jul 2015, pukul 10.23, Mike Willbanks <notifications@github.com> menulis:

Maks3w:
For me is not a BC break. `@internal` means for me is not part of the public API and must not be used.

Maks3w:
This can be merged in next maintenance release

PR #4, opened 2015-06-06
Added unit tests for the Metadata class

robertboloc:
This pull requests adds unit tests, with 100% coverage, for the Metadata class.

### DIFF
--- a/src/Metadata/Metadata.php
+++ b/src/Metadata/Metadata.php
@@ -41,7 +41,7 @@ class Metadata implements MetadataInterface
      *
      * @param  Adapter $adapter
      * @return Source\AbstractSource
-     * @throws \Exception
+     * @throws \Exception if the source cannot be created from the adapter
      */
     protected function createSourceFromAdapter(Adapter $adapter)
     {
--- a/src/Metadata/Metadata.php
+++ b/src/Metadata/Metadata.php
@@ -33,7 +33,7 @@ class Metadata implements MetadataInterface
     public function __construct(Adapter $adapter)
     {
         $this->adapter = $adapter;
-        $this->source = $this->createSourceFromAdapter($adapter);
+        $this->source  = $this->createSourceFromAdapter($adapter);
     }
 
     /**
@@ -41,6 +41,7 @@ class Metadata implements MetadataInterface
      *
      * @param  Adapter $adapter
      * @return Source\AbstractSource
+     * @throws \Exception
      */
     protected function createSourceFromAdapter(Adapter $adapter)
     {
--- a/test/Metadata/MetadataTest.php
+++ b/test/Metadata/MetadataTest.php
@@ -1,0 +1,182 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace ZendTest\Db\Metadata;
+
+use Zend\Db\Adapter\Platform\Mysql as MysqlPlatform;
+use Zend\Db\Adapter\Platform\SqlServer as SqlServerPlatform;
+use Zend\Db\Adapter\Platform\Sqlite as SqlitePlatform;
+use Zend\Db\Adapter\Platform\Postgresql as PostgresqlPlatform;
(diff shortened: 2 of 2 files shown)


src/Metadata/Metadata.php
```diff
@@ -41,6 +41,7 @@ public function __construct(Adapter $adapter)
      *
      * @param  Adapter $adapter
      * @return Source\AbstractSource
+     * @throws \Exception
```
Maks3w: Add description when this exception is thrown

test/Metadata/MetadataTest.php
```diff
@@ -0,0 +1,182 @@
...
+        );
+    }
+
+    /**
+     * @dataProvider createSourceFromAdapterDataProvider
+     *
+     * @param $platform
+     * @param $sourceExpectation
+     */
+    public function testCreateSourceFromAdapterCreatesPlatformSpecificSource($platform, $sourceExpectation)
```
Maks3w: This test is redundant with `testTheConstructorSetsTheAdaptersAndTheSource`

First one could use the data provider

test/Metadata/MetadataTest.php
```diff
@@ -0,0 +1,182 @@
...
+
+        $adapterReflectionProperty = new \ReflectionProperty($metadata, 'adapter');
+        $adapterReflectionProperty->setAccessible(true);
+        $adapterPropertyValue = $adapterReflectionProperty->getValue($metadata);
+
+        $sourceReflectionProperty = new \ReflectionProperty($metadata, 'source');
+        $sourceReflectionProperty->setAccessible(true);
+        $sourcePropertyValue = $sourceReflectionProperty->getValue($metadata);
+
+        $this->assertSame($adapterMock, $adapterPropertyValue);
```
Maks3w: Place this assertion after adapterProperyValue assignment

Maks3w: Better if use `$this->assertAttributeSame()` and remove all that reflection logic

test/Metadata/MetadataTest.php
```diff
@@ -0,0 +1,182 @@
...
+        $this->assertInstanceOf(
+            'Zend\Db\Metadata\Source\MysqlMetadata',
+            $sourcePropertyValue
+        );
+    }
+
+    /**
+     * @dataProvider createSourceFromAdapterDataProvider
+     *
+     * @param $platform
```
Maks3w: add param types or remove the tag

robertboloc:
@Maks3w I have update the pull request with your feedback from the code review, thanks!

test/Metadata/MetadataTest.php
```diff
@@ -0,0 +1,150 @@
...
+     * @param string           $sourceExpectation
+     */
+    public function testTheConstructorSetsTheAdaptersAndTheSource($platform, $sourceExpectation)
+    {
+        /** @var \Zend\Db\Adapter\Adapter $adapterMock */
+        $adapterMock = $this->getMockBuilder('Zend\Db\Adapter\Adapter')
+                            ->disableOriginalConstructor()
+                            ->getMock();
+
+        $adapterMock->expects($this->once())
```
Maks3w: Could you extract all the thing about `adapterMock` in a method called `createAdapterMockWithPlatform($platform)`

I think will improve test readiblity.

test/Metadata/MetadataTest.php
```diff
@@ -0,0 +1,150 @@
...
+            '',
+            false,
+            true,
+            true,
+            [$proxyMethod]
+        );
+
+        $sourceMock->expects($this->once())
+                   ->method($proxyMethod)
+                   ->willReturn($returnValue);
```
Maks3w: Could you test arguments received too?

Maks3w:
I start to think this class should be deprecated and only to have a factory class with a method for get the source from adapter

robertboloc:
I am not actually using the class. I just saw it had no coverage and I had some spare time so I decided to write tests for it. It does look like the only useful bit is the switch to get the source instance, as all the rest of the methods just proxy to the source.

I will leave the pull request as is for now, until a decision is made

Maks3w:
All the thing about to test proxy methods reveal a bad design flow.

mwillbanks:
@Maks3w that seems fair; however, i don't think it a bad idea to add the unit tests that have been add here.  Actually it is very well likely a great idea for the time being.  I would say we should merge the PR and then we can deprecate the other part in the future.

Maks3w:
I think I wrote a PoC with a more clean structure let me to dig on it

PR #5, opened 2015-06-09
Fix loading constraint data from Oracle

vixriihi:
This Fixes loading constraint data from Oracle database

### DIFF
--- a/src/Metadata/Source/OracleMetadata.php
+++ b/src/Metadata/Source/OracleMetadata.php
@@ -124,14 +124,14 @@ class OracleMetadata extends AbstractSource
                 AND cc2.position = cc1.position
 
             WHERE
-                ac.owner = :schema AND ac.table_name = :table
+                ac.owner = :ownername AND ac.table_name = :tablename
 
-            ORDER BY ac.constraint_name;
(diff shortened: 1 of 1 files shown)


Maks3w:
Need unit test

vixriihi:
I cannot add test before August

Maks3w:
I think this will closed then because there is no kind of explanation about what are you trying to fix

weierophinney:
@vixriihi Re-open or submit a new pull request when you have time to create a unit test.

vixriihi:
I'll add the unittests in another pr, but here is a small summary of what these chages would do. At the moment method loadConstraintData is not working at all in oracle and thats because there is [trailing semicolon](https://github.com/zendframework/zend-db/blob/master/src/Metadata/Source/OracleMetadata.php#L129) in the sql statement and because using your using reserved word [":table"](https://github.com/zendframework/zend-db/blob/master/src/Metadata/Source/OracleMetadata.php#L127) as variable

vixriihi:
Added the missing unittest and made a new PR https://github.com/zendframework/zend-db/pull/34

PR #15, opened 2015-07-08
Fixes #14 : Added Sql\Predicate\NotBetween

### LINKED ISSUE #14: Can you add a new class called "Zend\Db\Sql\Predicate\NotBetween"?
The NotBetween class is basically same as Between class, except its specification value is replaced with '%1$s NOT BETWEEN %2$s AND %3$s'. For example:
```php
<?php
/**
 * Zend Framework (http://framework.zend.com/)
 *
 * @link      http://github.com/zendframework/zf2 for the canonical source repository
 * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
 * @license   http://framework.zend.com/license/new-bsd New BSD License
 */

namespace Zend\Db\Sql\Predicate;

class NotBetween extends Between
{
    protected $specification = '%1$s NOT BETWEEN %2$s AND %3$s';
}
```
Without NotBetween class, my code will look like this:

```php
$between = new Between($identifier, "'$minValue'", "'$maxValue'");

if ('notBetween' === $predicate) {
     $between->setSpecification('%1$s NOT BETWEEN %2$s AND %3$s');
}

$predicate->addPredicate($between);
```

With NotBetween class, my code will look much more simple:

```php
$predicate->$predicate($identifier, "'$minValue'", "'$maxValue'");
```

Thanks for reading and I hope you may consider this!  =)

samsonasik:
Adds a `NotBetween` predicate:

```php
$sql->notBetween('count', 1, 5);
```

Where 'count' is the field identifier, 1 is the minimum (left side), and 5 is the maximum (right side).

### DIFF
--- a/src/Sql/Predicate/NotBetween.php
+++ b/src/Sql/Predicate/NotBetween.php
@@ -1,0 +1,15 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Db\Sql\Predicate;
+
+class NotBetween extends Between
+{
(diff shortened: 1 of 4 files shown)


weierophinney:
Merged to develop for release with 2.6.0.

Issue #18, opened 2015-07-10
Allow HydratingResultSet to accept a factory for an object prototype


waddedMeat:
I have a database table that makes use of a "type" column.  I would like to be able to pass in a factory as my `$objectPrototype` so my result set can contain objects from distinct classes instead of single class that contains methods that switch on "type"

This could be as simple as creating a `HydratingResultSet\PrototypeFactoryInterface` and changing the `HydratingResultSet::current()` method to check `instanceof` before just cloning the `$objectPrototype`

waddedMeat:
This is what I'm proposing (of course I would have unit tests)

```
diff --git a/src/ResultSet/HydratingResultSet.php b/src/ResultSet/HydratingResultSet.php
index a8e2013..ea8dcff 100644
--- a/src/ResultSet/HydratingResultSet.php
+++ b/src/ResultSet/HydratingResultSet.php
@@ -10,6 +10,7 @@
 namespace Zend\Db\ResultSet;
 
 use ArrayObject;
+use Zend\Db\ResultSet\HydratingResultSet\PrototypeFactoryInterface;
 use Zend\Stdlib\Hydrator\ArraySerializable;
 use Zend\Stdlib\Hydrator\HydratorInterface;
 
@@ -100,7 +101,13 @@ class HydratingResultSet extends AbstractResultSet
             return $this->buffer[$this->position];
         }
         $data = $this->dataSource->current();
-        $object = is_array($data) ? $this->hydrator->hydrate($data, clone $this->objectPrototype) : false;
+
+        if ($this->objectPrototype instanceof PrototypeFactoryInterface) {
+            $prototype = $this->objectPrototype->createPrototype($data);
+        } else {
+            $prototype = clone $this->objectPrototype;
+        }
+        $object = is_array($data) ? $this->hydrator->hydrate($data, $prototype) : false;
 
         if (is_array($this->buffer)) {
             $this->buffer[$this->position] = $object;
diff --git a/src/ResultSet/HydratingResultSet/PrototypeFactoryInterface.php b/src/ResultSet/HydratingResultSet/PrototypeFactoryInterface.php
new file mode 100644
index 0000000..5cd5b0c
--- /dev/null
+++ b/src/ResultSet/HydratingResultSet/PrototypeFactoryInterface.php
@@ -0,0 +1,23 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
[...]

PR #19, opened 2015-07-10
Allowed HydratingResultSet to accept a PrototypeFactoryInterface

waddedMeat:
This change allows an instance of a PrototypeFactoryInterface to be used
instead of an object prototype.  The factory will be passed the result
data and must return an object to be hydrated.

This would resolve Issue https://github.com/zendframework/zend-db/issues/18

### DIFF
--- a/test/ResultSet/HydratingResultSetIntegrationTest.php
+++ b/test/ResultSet/HydratingResultSetIntegrationTest.php
@@ -29,4 +29,27 @@ class HydratingResultSetIntegrationTest extends \PHPUnit_Framework_TestCase
         $obj2 = $hydratingRs->current();
         $this->assertSame($obj1, $obj2);
     }
+
+    /**
+     * @covers Zend\Db\ResultSet\HydratingResultSet::current
+     */
+    public function testCurrentUsesPrototypeFactory()
+    {
+        $dataSource = new \ArrayIterator([
+            ['id' => 1, 'name' => 'one'],
+            ['id' => 2, 'name' => 'two'],
+        ]);
+
+        $factory = $this->getMock('\Zend\Db\ResultSet\HydratingResultSet\PrototypeFactoryInterface');
+        $factory->expects($this->once())
+            ->method('createPrototype')
+            ->with($dataSource[0])
+            ->will($this->returnValue(new \ArrayObject));
+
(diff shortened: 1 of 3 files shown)


test/ResultSet/HydratingResultSetIntegrationTest.php
```diff
@@ -29,4 +29,27 @@ public function testCurrentWillReturnBufferedRow()
...
+        $dataSource = new \ArrayIterator([
+            ['id' => 1, 'name' => 'one'],
+            ['id' => 2, 'name' => 'two'],
+        ]);
+
+        $factory = $this->getMock('\Zend\Db\ResultSet\HydratingResultSet\PrototypeFactoryInterface');
+        $factory->expects($this->once())
+            ->method('createPrototype')
+            ->with($dataSource[0])
+            ->will($this->returnValue(new \ArrayObject));
```
weierophinney: Import `ArrayObject`, instead of using global qualifier. Same with `ArrayIterator` above.

test/ResultSet/HydratingResultSetIntegrationTest.php
```diff
@@ -29,4 +29,27 @@ public function testCurrentWillReturnBufferedRow()
...
+     * @covers Zend\Db\ResultSet\HydratingResultSet::current
+     */
+    public function testCurrentUsesPrototypeFactory()
+    {
+        $dataSource = new \ArrayIterator([
+            ['id' => 1, 'name' => 'one'],
+            ['id' => 2, 'name' => 'two'],
+        ]);
+
+        $factory = $this->getMock('\Zend\Db\ResultSet\HydratingResultSet\PrototypeFactoryInterface');
```
weierophinney: Class names inside strings are always considered globally qualified, and do not need the `\\` prefix. (Also seen in line 53.)

weierophinney:
I'm not convinced we need this; this sounds like a very application-specific use case, which would warrant a custom hydrator implementation.

As an example, you could overload `DelegatingHydrator::hydrate()` to introspect the `$data` argument to determine the object type to pass to its `getHydrator()` method. This seems like a cleaner solution than altering zend-db, as it separates the logic of object creation from the zend-db component — which was the original point of having the `HydratingResultSet`.

PR #25, opened 2015-07-21
add dependency from zendframework\zend-eventmanager

sebaks:
i try use zendframework/zend-db with --no-dev option on my production
```bash
$ mkdir 222
$ cd 222
$ composer require zendframework/zend-db
```
i create index.php:
```php
<?php

include "vendor/autoload.php";

$adapter = new Zend\Db\Adapter\Adapter([
    'driver' => 'Pdo_Pgsql',
    'database' => 'socdate',
    'username' => 'socdate',
    'password' => '111',
    'host' => 'localhost',
    'port' => 5432
]);

use Zend\Db\TableGateway\TableGateway;

$projectTable = new TableGateway('sd_roles', $adapter);
$rowset = $projectTable->select(array('status' => 1));
```
and run
```bash
$ php index.php
PHP Fatal error:  Interface 'Zend\EventManager\EventsCapableInterface' not found in /storage/proj/222/vendor/zendframework/zend-db/src/TableGateway/Feature/EventFeature.php on line 24
```

your require section must contain 'zendframework/zend-eventmanager'

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,10 +14,10 @@
     },
     "require": {
         "php": ">=5.5",
-        "zendframework/zend-stdlib": "~2.5"
+        "zendframework/zend-stdlib": "~2.5",
+        "zendframework/zend-eventmanager": "~2.5"
     },
     "require-dev": {
-        "zendframework/zend-eventmanager": "~2.5",
         "zendframework/zend-mvc": "~2.5",
         "zendframework/zend-servicemanager": "~2.5",
         "fabpot/php-cs-fixer": "1.7.*",


Maks3w:
Another approach is to replace constants references with constants values.

danielstqry:
This coupling was introduced with zendframework/zf2#7066. Can we resolve like it used to be, without always requiring EventManager?

gianarb:
zendframework/zf2#7066 increase the code quality.. IMO

We can add this dependency

mwillbanks:
Unfortunately this is an optional dependency depending on what component you use.  I've submitted the same exact PR a couple of months ago and it was determined that this is an optional dependency and does not require a hard dependency.  @weierophinney any other thoughts?

mwillbanks:
See PR #1

Maks3w:
As I said, another approach is to replace EventFeature constants with string values.

weierophinney:
Closing in favor of #41, which updates the library in a BC-safe way without needing to introduce the dependency.

Issue #26, opened 2015-07-21
False values for left outer and right outer join


tdutrion:
`Zend\Sql\Select` contains constant declaration for both left and right outer join as follow:

```
const JOIN_OUTER_RIGHT = 'outer right';
const JOIN_OUTER_LEFT  = 'outer left';
```

Unfortunately, as described in the [wikipedia article](https://en.wikipedia.org/wiki/Join_(SQL)#Left_outer_join) for instance, values should be respectively `left outer` and `right outer` instead of `outer left` and `outer right`.

weierophinney:
Closed with #27

PR #27, opened 2015-07-21
Fix values for left and right outer join #26

tdutrion:
Simple replacement of the values according to #26.

Theses values have been there since e6ad68226c0ee2d89181c2aecc92e7e56ca464e7 and names have been changed in ecdf83ed1002ad52c617ed024f6e00dc8d9c63b2. Changing back to `JOIN_LEFT_OUTER ` and `JOIN_RIGHT_OUTER ` would have been meaningful but not done here because of the BC break it involves.

### DIFF
--- a/src/Sql/Select.php
+++ b/src/Sql/Select.php
@@ -41,8 +41,8 @@ class Select extends AbstractPreparableSql
     const JOIN_OUTER = 'outer';
     const JOIN_LEFT = 'left';
     const JOIN_RIGHT = 'right';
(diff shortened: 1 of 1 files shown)


Maks3w:
This is a BC break anyway because use constant instead value is not mandatory.

You could do the following:

* Add two new constants with the suggested name
* Deprecate the old constants

weierophinney:
Merged to develop for release with 2.6.0.

PR #29, opened 2015-07-29
Allow null $parameters in Predicate::expression()

HillTravis:
Zend\Db\Sql\Predicate\Expression::__construct() allows nulls for the `$valueParameter`, but this method that passes the parameter through does not allow nulls, so you have to explicitly pass null. Instead, in Zend\Db\Sql\Predicate\Predicate::expression(), the `$parameters` parameter should allow nulls.

### DIFF
--- a/src/Sql/Predicate/Predicate.php
+++ b/src/Sql/Predicate/Predicate.php
@@ -242,7 +242,7 @@ class Predicate extends PredicateSet
      * @param $parameters
      * @return $this
      */
-    public function expression($expression, $parameters)
+    public function expression($expression, $parameters = null)
     {
         $this->addPredicate(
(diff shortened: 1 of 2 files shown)


mwillbanks:
@HillTravis can you create a unit test for this?  I have actually had this issue before which I could see this being useful.

HillTravis:
@mwillbanks I added a unit test. I'm not sure why the Travis CI build is failing or why coverage would have decreased with the ADDITION of a unit test, but I ran the tests in my fork and everything passed.

weierophinney:
@HillTravis The failure is a coding standards check... and one that we just fixed on another PR. I'll review now.

Thanks!

Issue #47, opened 2015-10-28
Improvement: Add unix_socket to dsn variables in PDO Adapter


KIVagant:
Sorry for my bad English.

Now, for connection to db through unix socket I need setup $dsn string manually as full string in config. It's not good, because I need to make some logic in configuration file.
I propose to add this lines to src/Adapter/Driver/Pdo/Connection.php after #239 row:

```
                    if (isset($unix_socket)) {
                        $dsn[] = "unix_socket={$unix_socket}";
                    }
```

PR #48, opened 2015-10-28
Adding unix_socket to dsn variables in PDO Adapter

KIVagant:
Issue #47 PR

### DIFF
--- a/src/Adapter/Driver/Pdo/Connection.php
+++ b/src/Adapter/Driver/Pdo/Connection.php
@@ -199,6 +199,9 @@ class Connection extends AbstractConnection
                 case 'charset':
                     $charset    = (string) $value;
                     break;
+                case 'unix_socket':
+                    $unix_socket = (string) $value;
+                    break;
                 case 'driver_options':
                 case 'options':
                     $value = (array) $value;
@@ -236,6 +239,9 @@ class Connection extends AbstractConnection
                     }
                     if (isset($charset) && $pdoDriver != 'pgsql') {
                         $dsn[] = "charset={$charset}";
+                    }
+                    if (isset($unix_socket)) {
+                        $dsn[] = "unix_socket={$unix_socket}";
                     }
                     break;
             }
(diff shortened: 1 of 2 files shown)


Maks3w:
Need test

KIVagant:
Ok, in progress...

KIVagant:
@Maks3w , I added test. Now look's better?

Maks3w:
Seems ok. I'll forward this to maintainers with more experience of this for evaluation.

KIVagant:
Hello again, @Maks3w . Who should to check the PR? Can I hope that it will be reviewed soon?

Maks3w:
ping @zendframework/community-review-team

src/Adapter/Driver/Pdo/Connection.php
```diff
@@ -237,6 +240,9 @@ public function connect()
                     if (isset($charset) && $pdoDriver != 'pgsql') {
                         $dsn[] = "charset={$charset}";
                     }
+                    if (isset($unix_socket)) {
```
mwillbanks: You *cannot* set the unix socket if a hostname or port is passed, this is really only relevant for mysql as well.

KIVagant: > You cannot set the unix socket if a hostname or port is passed
I think, PDO can resolve this issue without problems and zend-db code does not needed additional 'if-else' for this case. But, if you think another, I can add some checks for host and port variables.

> This is really only relevant for mysql as well.
In fact, [PDO_PGSQL DSN](http://php.net/manual/en/ref.pdo-pgsql.connection.php) can connect to unix socket too, but libpq-connect used "host" variable for path to socket.
What do you think I need to do? Add here if-condition for mysql-pdo only? Or add special condition for converting $unix_socket to $host variable for PDO_PGSQL?

mwillbanks: I think it would be best to add the if else statement since the manual does explicitly point it out for PDO.

mwillbanks: Actually, it might be better to throw an exception in that case; rather than leave it to undefined behavior.

KIVagant: ok, i'll try to resend PR tomorrow

KIVagant:
Sorry, I don't have enough time to finish and test this PR.

mwillbanks:
Hello @KIVagant I'll take care of the update in this case and update it based on what is needed.

KIVagant:
@mwillbanks, thank you.

PR #55, opened 2015-12-05
Implement FeatureSet canCallMagicCall and callMagicCall methods

settermjd:
Recently, I was attempting to call the lastSequenceId method of the
SequenceFeature, after adding the SequenceFeature to a TableGateway
object. However, on doing so, I discovered that both methods effectively
weren't implemented, despite what the manual alluded to. Given that I'm
adding this commit, along with the covering tests to implement the
functionality.

### DIFF
--- a/test/TableGateway/Feature/FeatureSetTest.php
+++ b/test/TableGateway/Feature/FeatureSetTest.php
@@ -14,6 +14,7 @@ use Zend\Db\TableGateway\Feature\MasterSlaveFeature;
 use Zend\Db\TableGateway\Feature\SequenceFeature;
 use Zend\Db\TableGateway\Feature\MetadataFeature;
 use Zend\Db\Metadata\Object\ConstraintObject;
+use \ReflectionClass;
 
 class FeatureSetTest extends \PHPUnit_Framework_TestCase
 {
@@ -156,7 +157,7 @@ class FeatureSetTest extends \PHPUnit_Framework_TestCase
             ->disableOriginalConstructor()
             ->getMock();
 
-        $reflectionClass = new \ReflectionClass('Zend\Db\TableGateway\AbstractTableGateway');
+        $reflectionClass = new ReflectionClass('Zend\Db\TableGateway\AbstractTableGateway');
         $reflectionProperty = $reflectionClass->getProperty('adapter');
         $reflectionProperty->setAccessible(true);
         $reflectionProperty->setValue($tableGatewayMock, $adapterMock);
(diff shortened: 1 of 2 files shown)


test/TableGateway/Feature/FeatureSetTest.php
```diff
@@ -77,4 +78,93 @@ public function testAddFeatureThatFeatureHasTableGatewayButFeatureSetDoesnotHas(
         $featureSet = new FeatureSet;
         $this->assertInstanceOf('Zend\Db\TableGateway\Feature\FeatureSet', $featureSet->addFeature($feature));
     }
+
+    /**
+     * @cover FeatureSet::canCallMagicCall
```
samsonasik: `@covers` instead of `@cover` ,  also, use fully namespaced class, `@covers Zend\Db\TableGateway\Feature\FeatureSet::canCallMagicCall`

settermjd: Thanks for that. I should have picked that up sooner. Perhaps I should go to bed as it's nearly 2am.

test/TableGateway/Feature/FeatureSetTest.php
```diff
@@ -77,4 +78,93 @@ public function testAddFeatureThatFeatureHasTableGatewayButFeatureSetDoesnotHas(
...
+        $adapterMock->expects($this->any())
+            ->method('getPlatform')->will($this->returnValue($platformMock));
+        $adapterMock->expects($this->any())
+            ->method('createStatement')->will($this->returnValue($statementMock));
+
+        $tableGatewayMock = $this->getMockBuilder('Zend\Db\TableGateway\AbstractTableGateway')
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $reflectionClass = new \ReflectionClass('Zend\Db\TableGateway\AbstractTableGateway');
```
samsonasik: you may use ReflectionProperty instead, also, move to use statement instead of adding "\" in front of it.

settermjd: Done in c34eaf5. Thanks for the suggestion.

test/TableGateway/Feature/FeatureSetTest.php
```diff
@@ -11,8 +11,10 @@
 
 use Zend\Db\TableGateway\Feature\FeatureSet;
 use Zend\Db\TableGateway\Feature\MasterSlaveFeature;
+use Zend\Db\TableGateway\Feature\SequenceFeature;
 use Zend\Db\TableGateway\Feature\MetadataFeature;
 use Zend\Db\Metadata\Object\ConstraintObject;
+use \ReflectionClass;
```
samsonasik: no more "\", also, sort it before "Z"

settermjd: done.

samsonasik:
ping @ralphschindler

settermjd:
Thanks @mwillbanks!

Issue #51, opened 2015-11-26
$rowset->current() return false instead of null when no data found


samsonasik:
For example, I have the following codes:
```php
use Zend\Db\TableGateway\TableGateway;
use Zend\Stdlib\Hydrator\ObjectProperty;
use Zend\Db\ResultSet\ResultSet;
use Zend\Db\Adapter\Adapter;

require 'vendor/autoload.php';

$adapter = new Adapter(array(
    'driver'   => 'pdo_mysql',
    'database' => 'db-album',
    'username' => 'root',
    'password' => ''
));


//using TableGateway instance
$sampleTable = new TableGateway('album', $adapter, null,new ResultSet());
//selecting...
$rowset = $sampleTable->select(array('artist' => 'Non-existent'));
$row = $rowset->current();

var_dump($row); // return false
```
While in code, it should be return `array|\ArrayObject|null` , see https://github.com/zendframework/zend-db/blob/master/src/ResultSet/ResultSet.php#L95

PR #56, opened 2015-12-07
data validation for AbstractResultSet:current

Keops92:
Fix for https://github.com/zendframework/zend-db/issues/51


Maks3w:
Fix the chmod

PR #59, opened 2015-12-09
Allow unix_socket parameter, throw exception if both hostname and unix_socket set

mwillbanks:
This PR supersedes PR #48 and fixes issue #47 

The main issue is that in the PDO MySQL adapter, if you want to connect via socket you need to pass the unix_socket parameter to PDO vs. a hostname parameter.  If both are set, it is ambiguous and is described in the PHP manual.  An exception will be thrown in this case.


weierophinney:
:+1: :shipit:

PR #65, opened 2015-12-15
Hotfix/zend db code comment correction

settermjd:
As I was stepping through `Zend\Db\Sql\Ddl\Index\Index` I noticed what seemed to be some discrepancies or inconsistencies in the constructor parameter documentation and argument signature. Given that, I'm submitting this PR. The specifics of what and why are contained in the commits.


settermjd:
Made a mistake with this one.

PR #66, opened 2015-12-15
Refactor Zend\Db\Sql\Ddl\Index\Index constructor

settermjd:
Refactored the name of the constructor's first argument and it's documentation to be in line with the `setColumns()` method which it's passed to. I came across this whilst debugging an `AlterTable` object which was attempting to add an Index to a PostgreSQL table.


settermjd:
Thanks @mwillbanks

PR #72, opened 2015-12-15
Changed Vagrantfile to reflect box rename to bento from chef

jchlu:
The `vagrant up` command was reporting it needed a name change from: 
Requested name: `chef/ubuntu-14.04` to Actual name: `bento/ubuntu-14.04`
This PR makes that change.


gianarb:
:+1:

PR #76, opened 2016-02-08
Refactor Insert::values() method to properly merge values with columns

settermjd:
When attempting to use both columns() and values() together to set the columns and value for an insert object, and when passing non-associative arrays to both, as seems logically correct to be able to do,
the generated SQL query has the numeric keys of the values array overwrite the columns set from the array passed to columns(). 

What I would expect is that the two were properly combined. This minor
fix does that, and leaves the remaining functionality in place. For example, I'd expect that if I had the following:

```php
$this->insert
    ->into('foo')
    ->columns(['col1', 'col2', 'col3'])
    ->values(['val1', 'val2', 'val3']);
```

That the generated SQL would be:

```sql
INSERT INTO foo ("col1", "col2", "col3") VALUES("val1", "val2", "val3");
```

Without this patch, the generated SQL would be:

```sql
INSERT INTO foo ("0", "1", "2") VALUES("val1", "val2", "val3");
```

### DIFF
--- a/src/Sql/Insert.php
+++ b/src/Sql/Insert.php
@@ -9,9 +9,9 @@
 
 namespace Zend\Db\Sql;
 
+use Zend\Db\Adapter\Driver\DriverInterface;
 use Zend\Db\Adapter\ParameterContainer;
 use Zend\Db\Adapter\Platform\PlatformInterface;
-use Zend\Db\Adapter\Driver\DriverInterface;
 
 class Insert extends AbstractPreparableSql
 {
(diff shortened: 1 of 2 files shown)


src/Sql/Insert.php
```diff
@@ -113,15 +114,33 @@ public function values($values, $flag = self::VALUES_SET)
...
             }
         }
         return $this;
     }
 
+
+    /**
+     * Simple test for an associative array
+     * @link http://stackoverflow.com/questions/173400/how-to-check-if-php-array-is-associative-or-sequential
+     * @param $arr
```
samsonasik: @param array $arr

src/Sql/Insert.php
```diff
@@ -113,15 +114,33 @@ public function values($values, $flag = self::VALUES_SET)
...
+     * @return bool
+     */
+    private function isAssocativeArray($arr)
```
samsonasik: `isAssocativeArray(array $arr)`

weierophinney:
Nice patch, @settermjd ! Thanks!

settermjd:
Not a problem at all @weierophinney. I'm aiming to be more active this year. Thanks for the feedback @samsonasik

PR #87, opened 2016-02-22
Extended test matrix for #85

weierophinney:
This pull request builds on #85, with the following changes:

- Allow zend-hydrator either `^1.1` or `^2.1`.
- Update the test matrix to allow testing zend-eventmanager v2 vs v3, and zend-hydrator v1.1 vs v2.1.
- Ensure that `AdapterServiceFactory` is both backwards compatible with v2 and forwards compatible with v3.


weierophinney:
Rebased and merged with dbf737e.

Issue #79, opened 2016-02-11
AbstractResultSet::current() change breaks consistency


Jakovitz:
AbstractResultSet::current() can buffer non-valid data.

Similar example as in #51:
```php
use Zend\Db\Adapter\Adapter;
use Zend\Db\ResultSet\ResultSet;
use Zend\Db\TableGateway\TableGateway;

require 'vendor/autoload.php';

$adapter = new Adapter(array(
    'driver'   => 'pdo_mysql',
    'database' => 'db-album',
    'username' => 'root',
    'password' => ''
));

//using TableGateway instance
$sampleTable = new TableGateway('album', $adapter, null,new ResultSet());
//selecting...
$rowset = $sampleTable->select(array('artist' => 'Non-existent'));
$rowset->buffer();
var_dump($rowset->current()); // null
var_dump($rowset->current()); // false
```

michalbundyra:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/133.

PR #89, opened 2016-03-01
Fix: Consistent AbstractResultSet::current #79

segrax:
#79 when calling 'current' multiple times, the return type changes. This is because 'false' is being stored in the buffer (result from current in the dataSource).

This is a result of the no-result return type change in f4ceeb689840d6bfbb99fa526e490d67dbb97c53 previously false was stored in the buffer, and also returned if $data was not an array.


samsonasik:
please provide tests.

weierophinney:
@segrax We need tests demonstrating the fix before we can merge this. Without tests, there's a strong possibility that a later change could re-introduce the issue.

Thanks in advance!

ezimuel:
Closing this PR since it's more than 1 year old.

Issue #7, opened 2015-06-11
Using JOINs in Zend\Db\Sql\Update


Koopzington:
You should be able to use JOINs in UPDATE statements the same way you do in SELECT statements.
If i'm not mistaken you should be able to use Expressions for values to make column names work in statements, so it should be quite simple to implement that.

metric152:
I'd love to see this
+1

PR #92, opened 2016-03-10
added join support in update

luciansabo:
Implements the feature requested here: https://github.com/zendframework/zend-db/issues/7

### DIFF
--- a/src/Sql/AbstractSql.php
+++ b/src/Sql/AbstractSql.php
@@ -78,7 +78,21 @@ abstract class AbstractSql implements SqlInterface
                 $sqls[$name] = $parameters[$name];
             }
         }
+
         return rtrim(implode(' ', $sqls), "\n ,");
+    }
+
+    /**
+     * Render table with alias in from/join parts
+     *
+     * @todo move TableIdentifier concatination here
+     * @param string $table
+     * @param string $alias
+     * @return string
+     */
+    protected function renderTable($table, $alias = null)
+    {
+        return $table . ($alias ? ' AS ' . $alias : '');
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/Sql/AbstractSql.php
```diff
@@ -291,6 +305,58 @@ protected function processSubSelect(
         return $decorator->buildSqlString($platform, $driver, $parameterContainer);
     }
 
+    public function processJoin(
```
weierophinney: Should be protected (per the other methods that are peers to it). I'll make that change on merge.

weierophinney: Also needs a docblock; I'll add that on merge.

src/TableGateway/AbstractTableGateway.php
```diff
@@ -326,7 +326,7 @@ protected function executeInsert(Insert $insert)
      * @param  string|array|\Closure $where
      * @return int
      */
-    public function update($set, $where = null)
+    public function update($set, $where = null, $joins = null)
```
weierophinney: Should be typehinted as `array`; will do that on merge.

src/TableGateway/AbstractTableGateway.php
```diff
@@ -337,6 +337,15 @@ public function update($set, $where = null)
         if ($where !== null) {
             $update->where($where);
         }
+
+        if ($joins !== null) {
+            if (is_array($joins)) {
+                foreach ($joins as $join) {
+                    $type = isset($join['type']) ? $join['type'] : null;
```
weierophinney: Should use `$update::JOIN_INNER` instead of `null` to get the default behavior. Will do that on merge.

PR #95, opened 2016-03-25
Overide quoteIdentifierInFragment function in IBMDB2 platform

phpdave:
The ```quoteIdentifierInFragment``` function in the ```AbstractPlatform``` class doesn't handle the number sign # (AKA pound, hash, octothorpe symbol) in a field.  IBM i DB2 tables commonly have # in the database field name.  The only change to the function is the regex.  I've added **#** 

'/([^0-9,a-z,A-Z$**#**_:])/i',

For example the field name WE_WEQSEQ# becomes WE_WEQSEQ**"#"** after running

```
$resultSet = $this->tableGateway->select(function (Select $select) use ($data) {
$select->where($data);
$select->order("WE_WEQSEQ# DESC");
});
```
And the database gives the error "Column or global variable WE_WEQSEQ"# not found. SQLCODE=-206"

## References to AbstractPlatform:

quoteIdentifierInFragment function
https://github.com/zendframework/zend-db/blob/master/src/Adapter/Platform/AbstractPlatform.php#L32

regex used in quoteIdentifierInFragment that picks the # to be quoted
https://github.com/zendframework/zend-db/blob/master/src/Adapter/Platform/AbstractPlatform.php#L45

## Regex verified by testing here
Live Test my regex '/([^0-9,a-z,A-Z$#_:])/i',:
https://regex101.com/r/oV2wL1/3

### DIFF
--- a/src/Adapter/Platform/IbmDb2.php
+++ b/src/Adapter/Platform/IbmDb2.php
@@ -40,7 +40,7 @@ class IbmDb2 extends AbstractPlatform
     {
         return 'IBM DB2';
     }
-    
+
     /**
      * {@inheritDoc}
      */
@@ -69,7 +69,7 @@ class IbmDb2 extends AbstractPlatform
         }
         return $identifier;
     }
-    
+
     /**
      * {@inheritDoc}
      */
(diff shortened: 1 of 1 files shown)


kingharrison:
Great stuff @phpdave !!

I have run into this particular issue quite often as the DB2 database I work on has many column names that include a # or $ because that type of naming is not considered wrong in the IBM i world.

alanseiden:
Yes, the pound or hash character (#) is frequently found in IBM i DB2 column names. Good idea to correct the regex.

alshayed:
We need this too, we also have many database tables with those characters. Thanks @phpdave!

Issue #98, opened 2016-04-13
Bug: select include all columns from joined tables (2.8.0 breaks)


belgattitude:
After upgrade from 2.7.0 to 2.8.0 :

When joining tables in a `Select` object, the `buildSqlString()` method includes all columns from joined tables (SQL_STAR) although they should be discarded by the third param in the `Select::join($name, $on, $columns = self::SQL_STAR, $type = self::JOIN_INNER)` method.

This bug may lead to really poor performance after upgrade as joined tables are often used to filter results or even worse disclosing information in case the query is used to generate data (api). This probably may be marked as security bugfix as well.

This bug have been tested with php 5.6.20 / 7, Mysql, Oracle, Sql92 platforms.

Take a look at this example 
_(notice the $columns=[] param in the join() method and see also the columns())_:

```php
$select = new \Zend\Db\Sql\Select();
$select->from('my_table')
       ->join('joined_table', 'my_table.id = joined_table.id', $columns=[])
       ->columns([
          'my_table_column',
       ]);
$sql_string = $this->sql->buildSqlString($select, $this->getAdapterForPlatform('Sql92'));
```
On 2.7.0 and before the resulting query string is correctly only producing one column:

```sql
SELECT 
    "my_table"."my_table_column" AS "my_table_column", 
FROM "my_table" 
INNER JOIN "joined_table" ON "my_table"."id" = "joined_table"."id"
```
After update to 2.8.0, the resulting query string **incorrectly** add all columns from the joined table. Take a look on the third line (joined_table.*)

```sql
SELECT 
    "my_table"."my_table_column" AS "my_table_column", 
    "joined_table".*
FROM "my_table" 
INNER JOIN "joined_table" ON "my_table"."id" = "joined_table"."id"
```

I'm currently looking to provide a P/R and add some specs into the SqlFunctionnalTests. Feel free to help.

belgattitude:
Pull request and possible fix on its way, see #99

weckx:
I can confirm this bug and it also breaks existing code, for example the query below that has two joins:

```
$select = $sql->select('news');
$select->join('company',` 'company.id = news.company_id');
$select->join('category', 'category.id = news.category_id');
```

If all the tables have an "id" field, In MySQL this will cause a "Duplicate column name", thus throwing an exception. 

For my case @belgattitude PR fixes the problem.

belgattitude:
Hi Felipe,

Good to know... 

[...]

PR #99, opened 2016-04-13
Fix Select joined table columns (2.8.0)

belgattitude:
Attempt to fix issue #98 

The current fix ensure the the columns of joined tables in a Select object can be excluded from sql string.

Since 2.8.0 the signature of the method join (see Join.php) has slightly changed. The default `Select::SQL_STAR` is now an array `[Select::SQL_STAR]`. 

Unfortunately the check for empty array with the ternary operator below always return SQL_STAR instead of "no columns" :

```php
        $this->joins[] = [
            'name'    => $name,
            'on'      => $on,
            'columns' => $columns ? $columns : [Select::SQL_STAR],
            'type'    => $type ? $type : Join::JOIN_INNER
        ];
```
This PR  changes the way it's tested, but I would love to get some feedback, especially for the strict equality operator... Looks it worked for me and phpunit but It may not be valid for every use case

```php
        if ($columns === [Select::SQL_STAR] || $columns == Select::SQL_STAR) {
            $columns = [Select::SQL_STAR];
        } elseif (!is_array($columns)) {
            $columns = [$columns];
        }
```

All the best

### DIFF
--- a/src/Sql/Join.php
+++ b/src/Sql/Join.php
@@ -125,7 +125,7 @@ class Join implements Iterator, Countable
                 sprintf("join() expects '%s' as a single element associative array", array_shift($name))
             );
         }
-        
+
         if ($columns === [Select::SQL_STAR] || $columns == Select::SQL_STAR) {
             $columns = [Select::SQL_STAR];
         } elseif (!is_array($columns)) {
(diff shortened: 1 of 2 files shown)


belgattitude:
Pull request closed and replaced by #100 

Just forgot to create a branch on my fork. Sorry

PR #100, opened 2016-04-13
Fix Select joined table columns (2.8.0) - new version

belgattitude:
Attempt to fix issue #98 

The current fix ensure the the columns of joined tables in a Select object can be excluded from sql string.

Since 2.8.0 the signature of the method join (see Join.php) has slightly changed. The default `Select::SQL_STAR` is now an array `[Select::SQL_STAR]`. 

Unfortunately the check for empty array with the ternary operator below always return SQL_STAR instead of "no columns" :

```php
        $this->joins[] = [
            'name'    => $name,
            'on'      => $on,
            'columns' => $columns ? $columns : [Select::SQL_STAR],
            'type'    => $type ? $type : Join::JOIN_INNER
        ];
```
This PR  changes the way it's tested, but I would love to get some feedback, especially for the strict equality operator... Looks it worked for me and phpunit but It may not be valid for every use case

```php
        if ($columns === [Select::SQL_STAR] || $columns == Select::SQL_STAR) {
            $columns = [Select::SQL_STAR];
        } elseif (!is_array($columns)) {
            $columns = [$columns];
        }
```

All the best

### DIFF
--- a/src/Sql/Join.php
+++ b/src/Sql/Join.php
@@ -125,7 +125,7 @@ class Join implements Iterator, Countable
                 sprintf("join() expects '%s' as a single element associative array", array_shift($name))
             );
         }
-        
+
         if ($columns === [Select::SQL_STAR] || $columns == Select::SQL_STAR) {
             $columns = [Select::SQL_STAR];
         } elseif (!is_array($columns)) {
@@ -137,8 +137,8 @@ class Join implements Iterator, Countable
             'on'      => $on,
             'columns' => $columns,
             'type'    => $type ? $type : Join::JOIN_INNER
-        ];        
-        
+        ];
+
         return $this;
     }
 
--- a/src/Sql/Join.php
+++ b/src/Sql/Join.php
@@ -126,10 +126,8 @@ class Join implements Iterator, Countable
             );
         }
 
-        if ($columns === [Select::SQL_STAR] || $columns == Select::SQL_STAR) {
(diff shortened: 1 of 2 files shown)


belgattitude:
Pull request updated...

Think we can get rid of my test for SQL_STAR (see the updated version in Join.php)... As far as I know, the only documented way to 'not' include columns from the joined table is to send an empty array. I also runned the [soluble-flexstore](https://github.com/belgattitude/soluble-flexstore) unit tests (heavily relies on zend-db) with success. 

See also improved functional tests (see SqlFunctionalTest.php)


PS:
I didn't test fully all possible 'undocumented' cases (like sending null or false instead of an empty array, some developers might have done it intuitively on older version). If anyone has ideas I tests different scenarios... Let me know

h3christophe:
Thank you !  I can't believe this bug went through like that - I just updated to 2.8.0 and it broke a lot of stuff in my site

weierophinney:
> I can't believe this bug went through like that 

I can; all tests passed locally when I merged, as well as on Travis. There simply weren't tests covering this particular behavior. Considering the enormity of SQL, there will always be edge cases we miss, as we cannot anticipate them all.

Thanks for the detailed reports, and thank you, @belgattitude, for the patch. I'll review and merge in the next day or so, and issue a release when I do.

belgattitude:
@weierophinney you're welcome, 

Thanks for all the good work and let me know in case i can help...

Seb

PR #107, opened 2016-04-24
AbstractResultSet implements JsonSerializable

vaclavvanik:
Since the requirements for PHP increased we could benefit from [JsonSerializable](http://php.net/manual/en/jsonserializable.jsonserialize.php).

### DIFF
--- a/src/ResultSet/AbstractResultSet.php
+++ b/src/ResultSet/AbstractResultSet.php
@@ -13,9 +13,10 @@ use ArrayIterator;
 use Countable;
 use Iterator;
 use IteratorAggregate;
+use JsonSerializable;
 use Zend\Db\Adapter\Driver\ResultInterface;
 
-abstract class AbstractResultSet implements Iterator, ResultSetInterface
+abstract class AbstractResultSet implements Iterator, JsonSerializable, ResultSetInterface
 {
     /**
      * if -1, datasource is already buffered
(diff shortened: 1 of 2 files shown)


Ocramius:
This seems to mix concerns, IMO. Serialization of resultsets is usually done via hydrators or via a simple `json_serialize(iterator_to_array($resultSet))`, which indicates the will to actually unroll the entire iterator.

Additionally, you may be running out of memory when treating an iterable object as an array, assuming it will just "fit".

I'll leave this decision to someone else, but I'm 👎 on this :-\

vaclavvanik:
@Ocramius thanks for your opinions. My idea to create this PR was that I can spare call of `json_encode(iterator_to_array($resultSet))`,  which still can be used. E.g.:

```
return new JsonResponse(['items' => $resultSet]);
```



Out of memory problem with encoding "big data" imo occurs anyway. Unfortunately `toArray()` has the same problem.

turrsis:
We can serialize to `XML` too, and, perhaps, to other formats.
Therefore this should be do via external objects.

mwillbanks:
I'm going to close this citing that we would be fixing concerns and the goal of the ResultSet is not to have this logic contained therein.  It is actually more beneficial to have the Hydrator do this work OR as @Ocramius has stated the simple difference.  When we then look at the comment about having XML and the several other formats this could start to cause a competing amount of code to come into the ResultSet that is not entirely necessary.

PR #112, opened 2016-05-08
Issue 111

jeanCarloMachado:
Test number of replacements on expression considers when same variable is used many times.


src/Sql/Expression.php
```diff
@@ -143,9 +143,13 @@ public function getExpressionData()
 
         // assign locally, escaping % signs
         $expression = str_replace(self::PLACEHOLDER, '%s', $expression, $count);
-        if ($count !== $parametersCount) {
+
+        //test number of replacements without considering same variable beign used many times first, which is
```
samsonasik: typo: s/beign/being

PR #116, opened 2016-06-21
Fix up the URLs used in the Coveralls badge

Koopzington:
[Coveralls](https://coveralls.io/github/zendframework/zend-db) says that it couldn't detect he badge in the Readme. Thought changing the URLs that the badge uses might fix this problem.


Koopzington:
Turns out the culprit was a deprecated option in the coveralls.yml.
Created a separate PR #117 
The currently used URLs for the badge (still) work but are not the same as the ones that the Coveralls site is recommending, so this PR has still a reason for existence ;)

Issue #142, opened 2016-06-28
SequenceFeature doesn't set lastInsertValue correctly


GeeH:
This issue has been moved from the `zendframework` repository as part of the bug migration program as outlined here - http://framework.zend.com/blog/2016-04-11-issue-closures.html

---

Original Issue: https://api.github.com/repos/zendframework/zendframework/issues/7689
User: @tiptone
Created On: 2016-03-17T19:23:50Z
Updated At: 2016-06-28T12:45:52Z
Body
In the file Module.php:

``` php
$dbAdapter = $sm->get('Zend\Db\Adapter\Adapter');
$featureSet = new FeatureSet();
$featureSet->addFeature(new Feature\SequenceFeature('ID','SEQNAME'));
$resultSetPrototype = new ResultSet();
$resultSetPrototype->setArrayObjectPrototype(new ModeledObject());
return new TableGateway('TABLENAME', $dbAdapter, $featureSet, $resultSetPrototype);
```

In the file ModelTable.php

``` php
$this->tableGateway->insert($data);
$id = $this->tableGateway->lastInsertValue;
```

Expected:
$id matches the value of TABLENAME.ID _after_ insert.

Result:
$id is the value of TABLENAME.ID _before_ the insert was done.  TABLENAME.ID is always +1 higher than the value returned by lastInsertValue.

---

PR #146, opened 2016-07-06
SequenceFeature doesn't set lastInsertValue correctly before the inse…

alexsds:
…rt was done
Fix for this issue: [#142](https://github.com/zendframework/zend-db/issues/142)
it is always +1 higher becouse lastInsertValue set from nextSequenceId()


alexsds:
@EvanDotPro please review

ezimuel:
We need also documentation on the Sequence feature, see #165

michalbundyra:
Rebased, updated and merged in 53fcd5f.

Thanks, @alexsds!

PR #153, opened 2016-07-25
Typo

gerardroche:


mwillbanks:
@gerardroche can you please add a unit test so this doesn't happen again :)

PR #156, opened 2016-08-26
Alter docs about Select and TableIdentifier usage

wandersonwhcr:
There's a typo in TableIdentifier usage in Select `from` method with alias.


ezimuel:
Already fixed.

PR #157, opened 2016-08-27
DDL with TableIdentifier

wandersonwhcr:
Enables usage of TableIdentifier in CreateTable, AlterTable and DropTable DDLs.


ezimuel:
Fixed conflicts, added phpdoc, and merged in develop. Thanks @wandersonwhcr!

PR #160, opened 2016-09-06
Feature/table gateway params

AdamiecRadek:
ping @weierophinney 
Added optional parameter for selectWith function where you can provid e additional parameters to execute function.
This will allow to user tablegateway features with flexibility of ordinary select

### DIFF
--- a/src/TableGateway/AbstractTableGateway.php
+++ b/src/TableGateway/AbstractTableGateway.php
@@ -201,7 +201,7 @@ abstract class AbstractTableGateway implements TableGatewayInterface
     /**
      * @param Select $select
      * @param array  $params - statement params
-     * 
+     *
      * @return ResultSetInterface
      * @throws \RuntimeException
      */
--- a/src/TableGateway/AbstractTableGateway.php
+++ b/src/TableGateway/AbstractTableGateway.php
@@ -25,25 +25,25 @@ use Zend\Db\TableGateway\Feature\EventFeatureEventsInterface;
 /**
  *
  * @property AdapterInterface $adapter
- * @property int $lastInsertValue
- * @property string $table
+ * @property int              $lastInsertValue
+ * @property string           $table
  */
 abstract class AbstractTableGateway implements TableGatewayInterface
 {
     /**
      * @var bool
      */
-    protected $isInitialized = false;
+    protected $isInitialized = FALSE;
 
     /**
      * @var AdapterInterface
      */
-    protected $adapter = null;
+    protected $adapter = NULL;
 
     /**
      * @var string|array|TableIdentifier
      */
-    protected $table = null;
+    protected $table = NULL;
 
     /**
      * @var array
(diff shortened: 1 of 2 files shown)


src/TableGateway/AbstractTableGateway.php
```diff
@@ -189,32 +191,35 @@ public function select($where = null)
...
     /**
      * @param Select $select
-     * @return null|ResultSetInterface
-     * @throws \RuntimeException
+     * @param array  $params - statement params
+     *
+     * @return null|\Zend\Db\ResultSet\ResultSetInterface
      */
-    public function selectWith(Select $select)
+    public function selectWith(Select $select, array $params = [])
```
alextech: Cannot make changes to functions without making corresponding changes to TableGatewayInterface signatures, where this is defined, and all other classes implementing it, if any.

src/TableGateway/AbstractTableGateway.php
```diff
@@ -25,25 +25,25 @@
...
+ * @property int              $lastInsertValue
+ * @property string           $table
  */
 abstract class AbstractTableGateway implements TableGatewayInterface
 {
     /**
      * @var bool
      */
-    protected $isInitialized = false;
+    protected $isInitialized = FALSE;
```
zf2timo: @AdamiecRadek i think this shouldn't be changed. Also the other occurs of `null` and `false`

froschdesign: @zf2timo 
Right!

@AdamiecRadek 
Please undo all code styling changes.

src/TableGateway/AbstractTableGateway.php
weierophinney: Not true. You can add optional arguments to implementations; it then becomes a matter of the user understanding whether or not they are using the interface signature, or the implementation signature. Considering the majority of TableGatewayInterface implementations start from this abstract class, it's fine — but we should likely add a todo item to the interface to add these new arguments in a new major version.

src/TableGateway/AbstractTableGateway.php
AdamiecRadek: @froschdesign undone

AGmakonts:
@weierophinney - Not directly related to this PR but I would like to know your opinion about this idea.

This PR was created to allow for usage of Table Gateway Features like master/slave with regular Select objects.

Why do those features need to be restricted to TG? Wouldn't it be better to have the option to set up additional features globally for an adapter? 

I know that master/slave feature requires the second adapter but possibly attaching slave adapter to the master is not such a bad idea.

Anyway..... I don't have any solutions to this yet but I this could be done I can start working on PR with those changes.

ezimuel:
@AdamiecRadek unit tests are missing, can you provide it? Thanks.

AdamiecRadek:
@ezimuel  Tests provided

michalbundyra:

michalbundyra:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/100.

Issue #67, opened 2015-12-15
Issues with DDL and PostgreSQL


settermjd:
I'm not sure if this is something which has already been covered, but recently I've been attempting to use the AlterTable and CreateTable classes with PostgreSQL and have been hitting an array of errors. These have been primarily caused by:

- Incorrect quoting of schema elements
- Incorrectly generated SQL

On investigating further, the support for PostgreSQL doesn't seem as robust as that for DB2, MySQL, Oracle or SQLServer. Am I missing something here? Or is it worth both implementing a set of Platform classes for PostgreSQL and spending time debugging and correcting the other issues which I've come across?

Would really appreciate feedback regarding this.

alextech:
I am trying to go through available abstraction utilities (data types, constraint types etc.) to check they work with PostgreSQL and update the newly created decorator as I find them.  Can you please share your usage examples that incorrectly generate SQL?

michalbundyra:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/135.

PR #163, opened 2016-09-06
Postgresql indexes

alextech:
Add support for Index creation with PostgreSQL platform. 
$table = CreateTable();
$table->addConstraint(new Index())
crashed with invalid syntax error.  Same for AlterTable.

Can be considered as first step towards cleaning up #67 
Continuing to work on finding the rest of the incompatibilities, but not sure if submit one large PR with full PgSQL support or do smaller more review manageable parts at a time. But then cannot make PRs depend on one another so, unless told otherwise will continue adding more commits here.

### DIFF
--- a/src/Sql/Ddl/Index/Index.php
+++ b/src/Sql/Ddl/Index/Index.php
@@ -26,7 +26,7 @@ class Index extends AbstractIndex
      * @param  null|string $name
      * @param array $lengths
      */
-    public function __construct($columns, $name = null, array $lengths = [])
+    public function __construct($columns = null, $name = null, array $lengths = [])
     {
         $this->setColumns($columns);
 
--- a/src/Sql/Platform/Platform.php
+++ b/src/Sql/Platform/Platform.php
@@ -13,6 +13,7 @@ use Zend\Db\Adapter\AdapterInterface;
 use Zend\Db\Adapter\Platform\PlatformInterface;
 use Zend\Db\Adapter\StatementContainerInterface;
 use Zend\Db\Sql\Exception;
+use Zend\Db\Sql\Platform\Postgresql;
 use Zend\Db\Sql\PreparableSqlInterface;
 use Zend\Db\Sql\SqlInterface;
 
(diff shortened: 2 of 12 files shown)


ezimuel:
@alextech thanks for this PR but it will require some time to review. In the meantime, because this is a new feature, can you send againt `develop` branch? I'm putting in zend-db 3.0.0 milestone.

alextech:
Changed.

rarog:
I checked the output of by generating a CreateTable for PostgreSQL.

The general syntax is nice and correct. Where it fails, is the correct escaping, if the escape character is inside column name.

The table name is escaped correctly, but the indices inside and outside of CREATE TABLE are going rogue.

```php
$table = new CreateTable('t\'e"s`t');
$table->addColumn(new BigInteger('t\'e"s`tCol'))
    ->addColumn(new BigInteger('t\'e"s`tCol2'))
    ->addConstraint(new Constraint\PrimaryKey('t\'e"s`tCol'));
    ->addConstraint(new Index('t\'e"s`tCol2', 't\'e"s`tIndex'));
```
Expected result:
```sql
CREATE TABLE "t'e""s`t" ( 
    "t'e""s`tCol" BIGINT NOT NULL,
    "t'e""s`tCol2" BIGINT NOT NULL,
    PRIMARY KEY ("t'e""s`tCol") 
); 
CREATE INDEX "t'e""s`tIndex" ON "t'e""s`tCol2");
```
Actual result:
```sql
CREATE TABLE "t'e""s`t" ( 
    "t""'""e""""""s""`""tCol" BIGINT NOT NULL,
    "t""'""e""""""s""`""tCol2" BIGINT NOT NULL,
    PRIMARY KEY ("t""'""e""""""s""`""tCol") 
); 
CREATE INDEX "t""'""e""""""s""`""tIndex" ON "t""'""e""""""s""`""t"("t""'""e""""""s""`""tCol2");
```

Instead of just doubling the escape character, there seems to be done some wrong and multiple escaping.

This behaviour is heavily bugged for the code base, too. For this I a separate bug report #267 and perhaps we could work on a fix together for all components.

alextech:
Thanks of great magnitude giving this a test after all this time. One thing to note, to make non-MySQL tests more realistic, try to always test with schema as part of the name. There are reports of escaper not working when using semantics outlined in last comment of https://github.com/zendframework/zend-db/issues/206 and fixed at https://github.com/zendframework/zend-db/pull/232

Then, there is also https://github.com/zendframework/zend-db/pull/233 to consider, which fixes inconsistency in applying identifier vs identifier chain. I do not *think* it relates to this issue, I doubt applying it will magically fix it, but sounds familiar. Will look into it.

rarog:
Good point, but so far the Ddl doesn't seem to offer a possibility to use schema as contrary to Select syntax, where TableIdentifier receives schema as 2nd parameter.

The other problem with the broken column and index names doesn't result in any of your parts but from `function processExpression` in `Zend\Db\Sql\AbstractSql`
```php
                } elseif ($type == ExpressionInterface::TYPE_IDENTIFIER) {
                    $values[$vIndex] = $platform->quoteIdentifierInFragment($value);
```

I'm about to comment on this in #267.

alextech:
Thanks @rarog I will see if table identifer can be used in DDL too.

michalbundyra:

michalbundyra:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/97.

PR #164, opened 2016-09-07
Added default 'db' configuration

weierophinney:
Defines a default adapter using an in-memory SQLite database. This is done to prevent failurs when adding zend-db to an application that retrieves the adapter via the container:

``` php
$container->get(\Zend\Db\Adapter\AdapterInterface::class)
```

This was observed when adding ZendDeveloperTools to a ZF3 application while also installing zend-db, but without first configuring zend-db.

Reported at zendframework/tutorials#57

### DIFF
--- a/src/ConfigProvider.php
+++ b/src/ConfigProvider.php
@@ -17,6 +17,7 @@ class ConfigProvider
     public function __invoke()
     {
         return [
+            'db'           => $this->getDefaultDbConfig(),
             'dependencies' => $this->getDependencyConfig(),
         ];
     }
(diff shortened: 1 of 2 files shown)


DragonBe:
When applying this patch the issue is resolved, so good choice.

I need to see if it's not causing issues when you're running this on a PHP installation without the sqlite extension. I think a **Required Prerequisites** section in the documentation could include it.

Or we can make the documentation adjust to the fact that when you install the Zend Toolbar to include the configuration.

samsonasik:
there is a PR at https://github.com/zendframework/ZendDeveloperTools/pull/218 to never call `$sm->get('Zend\Db\Adapter\AdapterInterface')` when 'db' index is undefined.

ezimuel:
@weierophinney is this PR still relevant?

DragonBe:
Not sure @ezimuel, the instances I've installed since the update after this bug was reported didn't have the issue. If necessary I can create a lab project to see if it's still relevant.

ezimuel:
@DragonBe yes, please. Thanks!

ezimuel:
@DragonBe, @weierophinney the PR https://github.com/zendframework/ZendDeveloperTools/pull/218 fixed the issue. I'm closing this PR.

PR #168, opened 2016-09-17
PhpDoc fixes

Koopzington:
Discovered some missing PhpDoc params, a param that didn't even exist and made sure PhpDoc reflects the method signature in AdapterAbstractServiceFactory


ezimuel:
@Koopzington thanks for your contribution!

PR #171, opened 2016-09-27
Fixed PHDoc

zf2timo:
Removed `null` as return value


ezimuel:
@zf2timo thanks for your contribution!

PR #162, opened 2016-09-06
Fixed invalid quoting for postgresql adapter

xorock:
As described in https://github.com/zendframework/zend-db/issues/161

PR #172, opened 2016-09-27
Fix nextval for sequence from outer schema

andrey-mokhov:
When sequence located in outer schema or sequenceName contain quote symbols, method nextSequenceId can't fetch next id, throw exception. For example:
```sql
SELECT nextval('"test.check_sequence_id_seq"');
```
This solution allow fix it

### DIFF
--- a/src/TableGateway/Feature/SequenceFeature.php
+++ b/src/TableGateway/Feature/SequenceFeature.php
@@ -12,6 +12,7 @@ namespace Zend\Db\TableGateway\Feature;
 use Zend\Db\Sql\Insert;
 use Zend\Db\Adapter\Driver\ResultInterface;
 use Zend\Db\Adapter\Driver\StatementInterface;
+use Zend\Db\Sql\TableIdentifier;
 
 class SequenceFeature extends AbstractFeature
 {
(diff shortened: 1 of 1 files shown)


alextech:
In any way related to https://github.com/zendframework/zend-db/pull/162 ?

andrey-mokhov:
#162 useful, when sequense name is known. But if not?
My solution is not using sequence name, which seted in constructor. 
It querying a next ID from DB by table name and primary key.

alextech:
This is a very good idea. Thanks for pointing out that flaw.

Does bring up a further problem, involving a choice for where does the sequence name come from in Table Gateway.  I had a client which did not have sequences always tied to a table in the form of a SERIAL clause, which would create implicit sequences that your changelist helps to locate. They prefer to manually request from the sequence each time and allocate it to column as they see fit. 

Secondly, sequences are not always used for PK columns.  There can be a sequence for primary key, sure, but some schema design patters use multiple sequences per table, sometimes one sequence across multiple tables.  Simplest example to think of is recept ID.  If have multiple tables for order types, each one will have one own sequence for order ID, then another shared sequence for receipt ID so all receipts are unique regardless of what order sub-system they come from.
Client I had used a similar technique for organizing tasks queues for scheduled workers.

So we have a problem where sometimes sequence is known, sometimes it is not known but as you point out is still expected to be there, and sometimes combination of both.  I will have to think about how to merge your solution with @xorock #162, including tests that this PR does not have. Unless you will beat me to it and have better solution, I may consider making a separate PR that puts the two of these together.

PR #177, opened 2016-10-11
Add $this->select to the output of getRawState() in Sql/Insert.php (Issue 175)

dmalouf:
Per Issue #175, this PR simply adds `$this->select` to the output of `getRawState()` (or `getRawState('select')`).  This is parallel to the way `Select::getRawState()` returns values, objects, or `null` - specifically, the way `Select` returns objects for: `having`, `table`, `where`.

That is, `getRawState()` does not have to only return scalars, arrays, or `null` - it can also return objects.  Therefore, this PR does not have `getRawState('select')` return the `getRawState()` of that `select` value (that is, this PR `$insert->getRawState('select')` is _not_ the same as `$insert->getRawState($select->getRawState())`).

### DIFF
--- a/src/Sql/Insert.php
+++ b/src/Sql/Insert.php
@@ -160,7 +160,8 @@ class Insert extends AbstractPreparableSql
         $rawState = [
             'table' => $this->table,
             'columns' => array_keys($this->columns),
-            'values' => array_values($this->columns)
+            'values' => array_values($this->columns),
+            'select' => $this->select
         ];
         return (isset($key) && array_key_exists($key, $rawState)) ? $rawState[$key] : $rawState;
     }
--- a/test/Sql/InsertTest.php
+++ b/test/Sql/InsertTest.php
@@ -125,6 +125,16 @@ class InsertTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * @covers Zend\Db\Sql\Insert::select
+     */
+    public function testSelectIncludedInRawState()
+    {
+        $select = new Select;
+        $this->insert->select($select);
+        $this->assertSame($select, $this->insert->getRawState('select'));
+    }
+
+    /**
      * @covers Zend\Db\Sql\Insert::prepareStatement
      */
     public function testPrepareStatement()


alextech (changes requested):

src/Sql/Insert.php
```diff
@@ -160,7 +160,8 @@ public function getRawState($key = null)
         $rawState = [
             'table' => $this->table,
             'columns' => array_keys($this->columns),
-            'values' => array_values($this->columns)
+            'values' => array_values($this->columns),
```
alextech: No, it does not prevent it. It actually helps.
This PR will not break any existing code.
However, I can see yet another bug report in sequence feature area in few weeks/month time with someone claiming a bug with a usecase I posted in the issue discussion.

I tend to get too verbose and sometimes my point get lost in the wall of text (to which this sentence ironically contributes) so will try to highlight the issue again.

To re-emphasize ,your code so far does not break anything, but is *incomplete*, if *filling out columns key is left out*. It can technically be accepted, but I personally do not feel it will be appropriate since that sequence thing is another bug report just waiting to happen and it can be prevented if this PR is expanded on to allow for a facility to fully inspect column values with SELECT the same way it is done with manual column specification.

dmalouf: @alextech - I'm slowly understanding more of what you're addressing.  Since I never do anything with Sequence stuff, I'm confident I do not fully grasp the situation fully at all (ha).

If I may offer a different perspective on multiple PRs for this `Insert.php` class: putting multiple Issues' needs into one PR can also be confusing 'down the road.'  If the value of the 'columns' key in `getRawState()` needs to have logic in it that is needed in the Sequence class(es), then it might be better to put such changes into the PR that uses such code changes.

I FULLY understand the dilemma of "PRs from different places depend on one another..."  What I'm not understanding (and this could very well have to do with my not knowing a thing about the Sequence stuff you're talking about) is how adding 'select' to `getRawState()` has any affect on Sequence other than actually making the Select object available?  If anything, it seems that adding 'select' makes this whole Insert object _more_ inspectable while keeping the Select object (and its columns) unique.

Perhaps you can point me to another `zend-db` class where a non-essential object's properties (values, column-names, something else) are 'added' to the essential properties of the containing class?  The other, main class that we use at work is the Select class which does _not_ pull data from classes it 'contains' (ex. Expression objects, TableIdentifier object, etc.) but instead simply returns the object-itself for the user to parse as-desired.

Or to look at it another way, if we change the content of `getRawState('columns')` in this PR, then would that not be a breaking change whereas adding the ability to inspect the Select object - even though this introduces more conditionals in situations like the Sequence stuff you mention - at least gives access to what is needed without breaking how the current method/contract works?

Again, if I am fully misunderstanding the situation you are describing, please just let me know and perhaps we can figure out a solution that accomplishes what you're seeing but - due to my lack of experience - doesn't require me to learn everything about the Sequence problems (grin).

alextech: _If anything, it seems that adding 'select' makes this whole Insert object more inspectable while keeping the Select object (and its columns) unique._
Yes, this is a major improvement already.  Many thanks for that :D :D

So part of me does agree that we can stop right here, and make an emphasis in documentation that whenever $insert->select() is used then getRawSate will not have 'columns' filled in and it will be responsibility of the API user to check for existence of 'select' key. If you go this route, I would recommend an appropriate documentation blob is added to this PR (https://github.com/zendframework/zend-db/blob/master/doc/book/sql.md#insert section) that usage of $insert->select() will make 'columns' not filled in and it will be up to the user to inspect select as you suggest for any potential conflicts manually.

Cant think of a good class to show at the moment :(

_how adding 'select' to getRawState() has any affect on Sequence other than actually making the Select object available_
This comes from API definition. Yes it may be obvious if someone reads through the source code of Zend DB prior to using it and realized that "oh, this code skips population of 'columns' in such and such conditions. Ok, I better accommodate for that". But then how many people read the source prior to usage? Then we have complains either ZF docs are incomplete or too hard to use because a developer who just wants to use the class will hit ctrl+space in editor, pick getRawState(), point it to 'columns' array and then what will he/she think when it is semi-randomly empty? Yet another bug report here. The effect it has on sequence, or any 3rd party lib inspecting the state, is something that would be covered by abstraction layer is not there.  Which leads to answer of 

_If we change the content of getRawState('columns') in this PR, then would that not be a breaking change whereas adding the ability to inspect the Select object - even though this introduces more conditionals in situations like the Sequence stuff you mention - at least gives access to what is needed without breaking how the current method/contract works_. 
If developers expect abstraction layer to do a job, in this case automatically supply list of columns that will be affected (or else whats the point of abstraction layer if I have to do things manually), then a case where that promise is broken and i have to fall back to manual inspections, **that** is breaking change to me.   If there is a common way to inspect Select object for columns, it should be included as a utility, as opposed to leaving it up to developers reinvent this part every time they call getRawState() (plus reading the source to understand why it does not follow the expectations).

I do not think you are misunderstanding as much as you say you do. (and this is not just about sequences anymore, that is just the first place I found a problem in, but API design in general) You are not giving yourself enough credit. You have good respectable arguments :) Hope I am not coming too strong against it and we are still on good discussion terms - you are giving me great thinking exercise :)

I feel we came to a standstill, both having valid arguments, and since I am not the framework's maintainer, I have no further say in right or wrong way. Will have to call to @weierophinney to be the judge.

dmalouf: Thank you for this great dialog, @alextech!  You are definitely **not** "coming too strong against" my thoughts on these topics - I have learned quite a bit plus you have given me a different perspective on writing APIs.  Very good stuff!

As I am new to being involved with large, public projects like ZF2, I was unsure what we do 'next' in our dialog.  I am glad you suggested we let @weierophinney be the judge - good idea!

Until the next PR :-)

michalbundyra:

michalbundyra:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/92.

Issue #35, opened 2015-09-07
Can't insert row when column name contains dash using PDO


dkmuir:
When there's a dash in the column name, inserts and updates fail.
See http://php.net/manual/en/pdo.prepared-statements.php#97162

The relevant code is:
https://github.com/zendframework/zend-db/blob/master/src/Adapter/Driver/Pdo/Pdo.php#L300

Is there a reason why this wouldn't work with positional placeholders instead?

alshayed:
I have a similar issue with my database - we commonly use the pound sign (#) in database field names which doesn't work since PDO only supports alphanumeric and underscore for placeholder/parameter names.

I don't know what downsides there would be to using positional placeholders, but I am working on a patch to filter the parameter names to only allowed characters. I've created a fork and have initial changes but it needs testing.

weierophinney:
Fixed with #224.

ezimuel:
For details, read all the comments of #224.

PR #178, opened 2016-10-12
Fix PDO parameter names

alshayed:
It appears that PDO only allows alphanumeric and underscore characters ([a-zA-Z0-9_]) for placeholder parameter names. However, the zend-db pdo driver just passes on anything without regard to those limitations. This patch addresses that by replacing any non-allowable characters with an underscore.

This is needed because various databases you might access through the PDO driver support more characters than that unquoted. For example MySQL also supports a dollar sign ($) and DB2 on the IBM i (I'm using this via PDO ODBC) supports the dollar sign and the hash/pound sign (#). (Side note - it works fine using the IbmDb2 driver which does positional parameters, but that is only available if you have that built in to PHP and are running it directly on the DB2 server or have purchased DB2 Connect)

I believe this addresses issue #35 though obviously not by changing to positional parameters as suggested there.

### DIFF
--- a/src/Adapter/Driver/Pdo/Pdo.php
+++ b/src/Adapter/Driver/Pdo/Pdo.php
@@ -298,7 +298,7 @@ class Pdo implements DriverInterface, DriverFeatureInterface, Profiler\ProfilerA
     public function formatParameterName($name, $type = null)
     {
         if ($type === null && !is_numeric($name) || $type == self::PARAMETERIZATION_NAMED) {
-            return ':' . $name;
+            return ':' . preg_replace('/[^a-zA-Z0-9_]/', '_', $name);
         }
 
         return '?';
--- a/src/Sql/Insert.php
+++ b/src/Sql/Insert.php
@@ -182,7 +182,7 @@ class Insert extends AbstractPreparableSql
             if (is_scalar($value) && $parameterContainer) {
                 $parameterName = $driver->formatParameterName($column);
                 $values[] = $parameterName;
-                if ($parameterName == '?') {
+                if ($driver->getPrepareType() == $driver::PARAMETERIZATION_POSITIONAL or $parameterName == '?') {
                     $parameterContainer->offsetSet($column, $value);
                 } else {
                     $parameterContainer->offsetSet(substr($parameterName, 1), $value);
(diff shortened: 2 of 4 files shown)


alshayed:
I'm not really sure how to add the tests to exercise this through the database. If somebody would be willing to turn this into the proper test code I would really appreciate it.

Here's how you can prove it manually with PDO/MySQL:

```SQL
CREATE TABLE test001 (
id integer not null auto_increment,
text_ char(10),
text$ char(10),

primary key (id)
);
```

Then example PHP in the root directory:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

$db_options = [
    'driver' => 'pdo',
    'dsn' => 'mysql:host=YOUR_DB_HOST;dbname=YOUR_DB_NAME;charset=utf8',
    'user' => 'YOUR_DB_USER',
    'pass' => 'YOUR_DB_PASS',
];

$db = new \Zend\Db\Adapter\Adapter($db_options);

$test001 = new \Zend\Db\TableGateway\TableGateway('test001', $db);
$test001->insert([
    'text_' => 'hi',
    'text$' => 'dude',
]);

$rowset = $test001->select();

$count = 0;
foreach ($rowset as $row) {
    $count++;
    echo json_encode($row) . PHP_EOL;
}

echo PHP_EOL . "found " . $count . " row(s)" . PHP_EOL;
```

So if you run it against the current master branch it fails with an exception, but my changes make it work as expected.

src/Adapter/Driver/Pdo/Pdo.php
```diff
@@ -298,7 +298,7 @@ public function getPrepareType()
     public function formatParameterName($name, $type = null)
     {
         if ($type === null && !is_numeric($name) || $type == self::PARAMETERIZATION_NAMED) {
-            return ':' . $name;
+            return ':' . preg_replace('/[^a-zA-Z0-9_]/', '_', $name);
```
ezimuel: This replace can produce conflicts in bindParam name. For instance, if you have 2 fields `text-` and `text$` they will be replaced with `text_` and `text_` both.

ezimuel:
Closing this in favor of #224

FlorentYaicene:
I am upgrading an app from php5.4 to php7.1 and I upgraded Zend db from 2.2 to 2.8.2 version.
Our database contains many column names with dash.

So, have you an approximative date of release fixing this issue ? 

I look forward to your response.

PR #179, opened 2016-10-12
Add consistency to method chaining DocBlocks

tdutrion:
As discussed in [fluent interface return type vs return self #7193](https://github.com/zendframework/zendframework/issues/7193), this changes remove occurrences of $this and self keyword from the DocBlocks.

### DIFF
--- a/src/Adapter/Adapter.php
+++ b/src/Adapter/Adapter.php
@@ -101,7 +101,7 @@ class Adapter implements AdapterInterface, Profiler\ProfilerAwareInterface
 
     /**
      * @param Profiler\ProfilerInterface $profiler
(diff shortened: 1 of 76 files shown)


tdutrion:
Updated to latest develop + changed to use class names in docblocks

ezimuel:
@tdutrion I'm sorry to manage this PR after 6 months... as discussed [here](https://github.com/zendframework/zendframework/issues/7193) we should use **self** if the same instance is returned and the instance **class name** otherwise. 
Can you review your PR? I see usage of class name instead of self. Thanks!

ezimuel:
@tdutrion thanks for your contribution!

PR #184, opened 2016-10-28
Critical Sqlsrv fix for PHP 7

marcelto:
Replaced literal empty string with references to new
parameterReferenceValues array property.  Updated values were not carrying through to the database.

### DIFF
--- a/src/Adapter/Driver/Sqlsrv/Statement.php
+++ b/src/Adapter/Driver/Sqlsrv/Statement.php
@@ -74,6 +74,11 @@ class Statement implements StatementInterface, Profiler\ProfilerAwareInterface
     protected $prepareOptions = [];
 
     /**
+     * @var array
+     */
+    protected $parameterReferenceValues = [];
+
+    /**
      * Set driver
      *
      * @param  Sqlsrv $driver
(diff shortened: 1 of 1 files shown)


GeeH:
@weierophinney do you have any comments on this please, it looks like a fairly simple fix for an urgent bug but I don't understand the context enough to comment

Ocramius:
Missing test case. Am I right to assume that this looks like https://github.com/doctrine/dbal/pull/2434 ?

ezimuel:
@marcelto can you provide a unit test for this PR? Thanks!

marcelto:
@ezimuel, I would but I don't think a unit test will cover the issue.  Everything looks perfect until the sqlsrv extension actually executes the statement and then database record contains nothing but empty values.

ezimuel:
@marcelto we can write an integration test, like [this](https://github.com/zendframework/zend-db/blob/master/test/Adapter/Driver/Sqlsrv/ConnectionIntegrationTest.php).
Moreover, I updated the [Vagrantfile](https://github.com/zendframework/zend-db/blob/master/Vagrantfile) to install SQL Server on Linux for testing.

marcelto:
@ezimuel ok I'll put something together.  Is there a table definition you'd like me to use?

ezimuel:
@marcelto you can use whatever table structure you want. The connection config is reported in [phpunit.xml.dist](https://github.com/zendframework/zend-db/blob/master/phpunit.xml.dist#L48-L52).

marcelto:
@ezimuel I'm thinking now that this could covered by a unit test.  Please review the gist and let me know if you agree.

[https://gist.github.com/marcelto/3ce68dfe47b9531a9d27a17775744a90](url)

marcelto:
Fixed link to gist

marcelto:
Closing since this is now a duplicate of https://github.com/zendframework/zend-db/pull/253.

Issue #186, opened 2016-11-09
[WIP][RFC]sequences support


alextech:
- [x] Explicit sequence with string or array identifier
- [x] Multiple sequences per table (would love ideas for improvement)
- [x] Implicit sequence with column name only for `SERIAL` pseudo-type
   - [x] Generate based on PostgreSQL concatenation rule
   - [x] Query database metadata if identifier ends up being longer than 63 bytes
- [x] Add `SERIAL` pseudo-type to column creation DDL to compliment idea of implicit sequence at DB creation/migrations level.
- [ ] Update `Insert` and `Update` statements to filter out or generate values based on the changes
   - [ ] Integrate with #177 
- [ ] Consider making `SequenceFeature` independent from `TableGateway` since (1) it does not necesseirly tie to a specific table, (2) fix `RowGateway` limitation
- [ ] Update Metadata TableGateway feature to recognize SERIAL columns and add SequenceFeature instances
- [ ] Add MS SQL Server 2016 support
- [ ] Stretch goal: add Oracle 12c implicit sequence support

No BC breaks.

Thought process history
--------------------------
While reviewing https://github.com/zendframework/zend-db/pull/172, helping with https://github.com/zendframework/zend-db/pull/162, and fresh out of a heavy database driven enterprise platform contract, noticed how deficient current sequence support is plus misleading function signatures.  This is an important feature because, better than auto_increment in PostgreSQL, the only way to increment Oracle, and now Microsoft brought it in due to popular demand in SQL Server 2016 so I feel it needs detailed look.

Current situation is that:

- TableGateway can only deal with a single sequence instance primarily for primary key usage, despite accepting unrestricted number of features.
- As pointed out by @andrey-mokhov only supports explicit sequence naming. In some engines like Oracle, MSSQL2016 this is fine, but PostgreSQL supports auto-generated sequence names with SERIAL which is a problem.
- Discovered at https://github.com/zendframework/zend-db/issues/191 that RowGateway is also incapable of handling sequences. Limitation is in fact that RowGateway does not link to any table gateway. Easy to add as an optional dependency by making constructor support TableGateway along with string and TableIdentifier *OR* add addition of features like other db classes, but hard in that INSERT operation on new record cannot be engine augnostic. May need to explicitly check for existence of that feature even if it breaks good OOP practice (or better suggestion).

What I have seen in the field:

[...]

PR #187, opened 2016-11-09
#186 [RFC][WIP] Sequence Support

alextech:
#186 [RFC][WIP] Sequence Support

Update 24th December 2016:
---------------------------------
Supersedes, combines PR #162 and #172. Addresses issues brought up by @andrey-mokhov about implicit sequences. To avoid extra query, first constructing name using same naming rule as PostgreSQL and when on rare occasions exceed length, query database. Improve on original solution by using prepared statement in order to cache queries instead of recompiling what is a very common statement with only identifier name variation.
Combines with @xorock solution of specifying schema together with sequence name. Parametrized statement should also deal with error prone manual value escaping.

FeatureSet filter
-------------------------------

First, need to adjust FeatureSet to handle more than one instance of a type to categorize _callMagicCall()_ properly.

This should help with the fact that even though TableGateway accepts unrestricted array of features, if more than one feature of the same type is added, only first one will be processed. Useful for sequences or anyone else who may have run into this.

What I am not fully happy with, could use verification on before I go too far into this:

1. Features of same type now need to have a filter of the target method to ensure they are indeed the target recipient. https://github.com/zendframework/zend-db/compare/master...alextech:sequences?expand=1#diff-85fe938f5a6f87443ad2578472eed9e0R112
This is needed so can have `$tableGateway->lastSequenceId('implicit_by_column_name');` or `$tableGateway->lastSequenceId('explicit_sequence_name');` calls to be delegated to appropriate sequence instance out of the array of them.  Not the cleanest looking solution.

2. Is categorization by class name the best way to go? Is https://github.com/zendframework/zend-db/compare/master...alextech:sequences?expand=1#diff-a46ec23198cadca611ac2ff22cbe055dR156 the best way to loop through them? Maybe to address first point should use combination of feature class name and some unique identifier like name. But then not all features will have an obvious identifier.

### DIFF
--- a/doc/book/table-gateway.md
+++ b/doc/book/table-gateway.md
@@ -293,11 +293,12 @@ There are a number of features built-in and shipped with zend-db:
   to run an extra metadata query on every `TableGateway` object construction. If that is the case,
   take note of what PostgreSQL created using
   ```sql
-    SELECT 'column_default' FROM information_schema.columns WHERE
-    table_schema = 'public'
-    AND table_name = 'artist'
-    AND column_name = 'id'
-    AND column_default LIKE 'nextval%';
+    SELECT 'column_default'
+    FROM information_schema.columns 
+    WHERE table_schema = 'public'
+      AND table_name = 'artist'
+      AND column_name = 'id'
+      AND column_default LIKE 'nextval%';
   ```
   
   take note of what `nextval` is reading from, and add it to `SequenceFeature` constructor.
--- a/doc/book/table-gateway.md
+++ b/doc/book/table-gateway.md
@@ -286,19 +286,14 @@ There are a number of features built-in and shipped with zend-db:
   
   With second parameter left null, `TableGateway` will generate sequence name based on same rule
   PostgreSQL uses (*tablename_columnname_seq* but if the resultant name is determined to be greater than 
-  63 characters, an additional query will be made to database schema to find what PostgreSQL has created 
+  63 bytes, an additional query will be made to database schema to find what PostgreSQL has created 
   instead, since transaction rules are more complex.
   
   This is important to know if you have long table and column names, and do not want
   to run an extra metadata query on every `TableGateway` object construction. If that is the case,
   take note of what PostgreSQL created using
   ```sql
-    SELECT 'column_default'
-    FROM information_schema.columns 
-    WHERE table_schema = 'public'
-      AND table_name = 'artist'
-      AND column_name = 'id'
-      AND column_default LIKE 'nextval%';
+    pg_get_serial_sequence('artist', 'id');
   ```
   
   take note of what `nextval` is reading from, and add it to `SequenceFeature` constructor.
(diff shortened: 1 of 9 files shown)


ezimuel:
Unfortunately, I've to close this PR due to inactivity.

alextech:
@ezimuel I was waiting on some assistance about difficult architecture decisions outline in https://github.com/zendframework/zend-db/issues/186 I got stuck not knowing how to go about those conflicts, not knowing what more experienced maintainers would do in that case, but never got reply :(

froschdesign:
@ezimuel 
We should reopen this PR, because @alextech does a good job here. Only some feedback is needed! Maybe you can review this PR.

froschdesign:
@ezimuel 
Inactivity lies on our side!

@alextech 
Would you complete this PR?

ezimuel:
@alextech ok, I'm re-opening this PR. I'll try to schedule some time for review it. It's a lot of code :(

alextech:
@froschdesign I will put the effort in! Thank you! It'll be challenging to rethink something this complicated a year later so I will take some days to catch up to my older self, but an opportunity to make Zend actually work I do not want to miss.  If this PR is closed, then several bug reports and PRs from others about sequences will have to be ignored: they are very deep rabbit holes.

@ezimuel It is a lot of code indeed :( For now, would you be able to at least go through conversations I had linked from #186? Especially the choice to be made in last comment of that issue, while keeping in mind last conversation at #162 where I mention that PDO driver solves _some_ of the problems out of the box. (as an extra, may glance through #177)

Also, if I can assume #233 will be merged in, that will help, because it resolves quoting duplication logic.

alextech:
@ezimuel Moving some of current questions I have from original bug report here so its easier for you to follow. Thank you for the blog post! Was motivational. Not pressuring you to answer these right away 

Original code had switch/case statement per platform. Thats what bug reports focused on so I followed same pattern: https://github.com/zendframework/zend-db/pull/187/files#diff-85fe938f5a6f87443ad2578472eed9e0R166

But I just noticed there is duplication of that lastSequenceId() query in Connection class. Seems like a great benefit because do not need nasty switch($platform), and if anyone relied on similarly broken code in Connection too, all is fixed in one place.

https://github.com/zendframework/zend-db/blob/master/src/Adapter/Driver/Pgsql/Connection.php#L267

https://github.com/zendframework/zend-db/blob/master/src/Adapter/Driver/Pdo/Connection.php#L419

There are some concerns, however. What about nextSequenceId()? Connection interface does not have signature for it the way getLastGeneratedValue() is and rightfully so, I believe, should not be added. I could add nextSequenceId() to connection classes without adding to interface and assume dynamic typing will handle it, but that looks dirty. But code duplication looks dirty too, and so does inconsistency of doing switch/case for one, and call to adapter for the other method.

Another possible downside of using adapter connection, is getLastGeneratedValue() does not share same syntax for SqlServer 2016 where its one command for IDENTITY column but a different one for SEQUENCE object/column. (not a problem for postgresql or oracle up to 12c because sequence is the only way to do it)

Finally, how do I go about deducing implicit sequence names which too is a platform specific action? My getSequenceName() is awefully postgresql specific. I am not sure if decorator is possible to do for TableGateway features.

weierophinney:

michalbundyra:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/88.

Issue #185, opened 2016-11-08
Pdo driver is passing platform & platform_options to PDO


alshayed:
I'm using pdo_ibm and if you specify the platform & platform_options keys on the zend-db Adapter constructor, it ultimately passes them on to the PDO constructor. Unfortunately with pdo_ibm it seems to silently ignore any other options after that, so this is bad. The good news is that if I put the platform & platform_options keys last on the Adapter constructor array that lets me work around the issue.

I think in src/Adapter/Driver/Pdo/Connection.php it should ignore the platform & platform_options keys in connect() when it's building the new driver options array. Ie just before line 210 (the default case) it should have case statements for platform & platform_options with only a break so it ignores them.

I'd be happy to submit a pull request for this if people agree - just want to make sure that this is the "right way" to do it. I have not tried to reproduce this with other pdo drivers, I'm not sure if it's an issue with the base PHP PDO implementation or specific to pdo_ibm.

Thanks!
Brandon

ezimuel:
Closing this because is not an issue, as reported in #189.

PR #189, opened 2016-11-12
Pdo/Connection::connect() should not pass platform keys to \PDO constructor

alshayed:
This is to resolve issue #185. With the pdo_ibm driver when passed the platform & platform_options keys it stops processing anything after that. Might be an issue with other drivers but I'm not sure. Since the platform & platform_options key are specifically related to zend-db I don't see any reason we'd ever want them passed to \PDO. Also even if the base \PDO driver needed those keys for some reason they could be put in a driver_options array which is what I am personally already doing anyway.

### DIFF
--- a/src/Adapter/Driver/Pdo/Connection.php
+++ b/src/Adapter/Driver/Pdo/Connection.php
@@ -207,6 +207,9 @@ class Connection extends AbstractConnection
                     $value = (array) $value;
                     $options = array_diff_key($options, $value) + $value;
                     break;
+                case 'platform':
+                case 'platform_options':
+                    break;
                 default:
                     $options[$key] = $value;
                     break;


ezimuel:
@alshayed this is not an issue of the source code, maybe a missing part in the IBM Db2 documentation.
The `platform` and `platform_options` are not supported Adapters keys, see [here](https://zendframework.github.io/zend-db/adapter/#creating-an-adapter-using-configuration). As you mentioned in #185 these values can be specified as optional parameters.

We should improve the documentation of IBM Db2 adapter. @alshayed would you like provide a PR to add it in the documentation?

alshayed:
@ezimuel if that's unsupported then why is it handled in the Adapter source code [here](https://github.com/zendframework/zend-db/blob/master/src/Adapter/Adapter.php#L325)?

Also I'm not entirely clear on exactly what changes to the documentation you are suggesting be made. I'll have to look at that but I think the question of this being supported or not should be addressed first.

alshayed:
@ezimuel here's the original [commit](https://github.com/zendframework/zend-db/commit/95e0fb8e492171a2324bd1974711072d1289fbc3) that appears to have added the platform_options functionality.

ezimuel:
@alshayed Ops, you right! I re-open the pull request to investigate more. I'm sorry for my oversight.

alshayed:
An arguably "better" solution than my PR would be to split out the platform related keys in the Adapter constructor, but it might have a higher risk of causing other people issues than my PR. Especially since I don't have an Oracle database to test against, my change was something I could better test on my own and am actually using in production.

However that does bring up the question of unit tests, which I really don't know how to handle something like this in the test suite. I'm happy to help with some guidance, or I think I've got this set to accept changes from yourself too.

ezimuel:
@alshayed can you provide an extract of your code where you use pdo_ibm with driver_options? Thanks.

alshayed:
No problem, here's a simple version of what I'm doing on the server. Looking at the documentation you linked earlier this also doesn't appear to be listed as an allowable key, but is handled in several of the drivers.

```php
    use Zend\Db\Adapter\Adapter;

    $settings = [
        'db' => [
            'dsn' => 'ibm:SYSTEMNAME',
            'driver' => 'pdo',
            'driver_options' => [
                PDO::ATTR_PERSISTENT => true,
                PDO::ATTR_AUTOCOMMIT => true,
                PDO::I5_ATTR_DBC_SYS_NAMING => true,
                PDO::I5_ATTR_DBC_CURLIB => '',
                PDO::I5_ATTR_DBC_LIBL => 'SCHEMA1 SCHEMA2 SCHEMA3',
            ],
            'username' => '__USER__',
            'password' => '__PASS__',
            'platform' => 'IbmDb2',
            'platform_options' => ['quote_identifiers' => false],
        ],
    ];

    $db_adapter = new Adapter($settings['db']);
```

The main reason I'm using PDO in general vs the main IbmDb2 driver is because with PDO on my development PC I can use ODBC to connect to the database and have it still work mostly the same - certainly more so than using the IbmDb2 driver. (I can't recall off hand exactly the differences I ran into but there were some, I think some involved how it returned field names from the database)

When I'm using pdo_odbc on my PC I change the dsn to `odbc:connection_name` and then don't need the driver_options because that's all handled in the ODBC connection setup. I think there's a way I could do everything in the PHP code but I haven't experimented with that yet. My understanding is that it would be part of the dsn string.

ezimuel:
@alshayed the platform and platform_options are used only by IbmDb2 and Oralce adapter, not PDO.
As commented previously, this is not an issue the code but a documentation lack. I think we need to provide examples of db connection using `Zend\Db\Adapter\Adapter` for all the supported drivers.
We can add this notes [here](https://github.com/zendframework/zend-db/blob/master/doc/book/adapter.md#creating-an-adapter-using-configuration), you can see the rendered version [online](https://zendframework.github.io/zend-db/adapter/#creating-an-adapter-using-configuration).
Would you like to add a section for `pdo_ibm` and `IbmDb2`? Thanks!

alshayed:
@ezimuel I apologize for my apparent failure to fully explain what I see the issue as here. I will try one last time and then drop it if you still disagree that there should be any fix made. Also I realize you likely don't have the ability to test this easily, but since you work at Zend I suspect Rod or Shlomo might be able to help you. I can even file a support ticket with them if that would help in any way.

First off I agree that the `platform` and `platform_options` keys are not being "used" by PDO. The problem is that PDO doesn't like receiving unsupported keys and ignores all other keys following them. So in my example above - since I have those keys last it all works out fine in the end. But if I put them first in the array then PDO will not get set up properly. BTW last I knew this requires PHP 5.6 because Zend & IBM still haven't fixed the pdo_ibm driver to work properly on PHP 7.x on the IBM i platform.

So all I'm asking for here is a small fix to prevent PDO from being given keys on the options parameter that it doesn't support. This would be a good improvement for two reasons: first to help new people getting started so they don't have a big problem for the sole reason of having the keys in the wrong order and secondly because it's just good practice to not give known unsupported data to something like PDO.

Below are two examples - one working and one not, the only difference in the two is what order the zend-db settings keys are in.

Working example:

```php
<?php

use Zend\Db\Adapter\Adapter;

require __DIR__ . '/vendor/autoload.php';

$settings = [
    'db' => [
        'dsn' => 'ibm:DB_NAME',
        'driver' => 'pdo',
        'driver_options' => [
            PDO::ATTR_PERSISTENT => true,
            PDO::ATTR_AUTOCOMMIT => true,
            PDO::I5_ATTR_DBC_SYS_NAMING => true,
            PDO::I5_ATTR_DBC_CURLIB => '',
            PDO::I5_ATTR_DBC_LIBL => 'SCHEMA1 SCHEMA2 SCHEMA3',
        ],
        'username' => '__USER__',
        'password' => '__PASS__',
        'platform' => 'IbmDb2',
        'platform_options' => ['quote_identifiers' => false],
    ],
];

$db_adapter = new Adapter($settings['db']);

$table_gateway = new \Zend\Db\TableGateway\TableGateway('SCGENLCK', $db_adapter);

$rowset = $table_gateway->select();

$count = 0;
foreach ($rowset as $row) {
    $count++;
}

echo PHP_EOL . "found " . $count . " row(s)" . PHP_EOL;
```

result:
```
$ php pdo_test.php

found 23 row(s)
```

Non-working example:

```php
<?php

use Zend\Db\Adapter\Adapter;

require __DIR__ . '/vendor/autoload.php';

$settings = [
    'db' => [
        'platform' => 'IbmDb2',
        'platform_options' => ['quote_identifiers' => false],
        'dsn' => 'ibm:DB_NAME',
        'driver' => 'pdo',
        'driver_options' => [
            PDO::ATTR_PERSISTENT => true,
            PDO::ATTR_AUTOCOMMIT => true,
            PDO::I5_ATTR_DBC_SYS_NAMING => true,
            PDO::I5_ATTR_DBC_CURLIB => '',
            PDO::I5_ATTR_DBC_LIBL => 'SCHEMA1 SCHEMA2 SCHEMA3',
        ],
        'username' => '__USER__',
        'password' => '__PASS__',
    ],
];

$db_adapter = new Adapter($settings['db']);

$table_gateway = new \Zend\Db\TableGateway\TableGateway('SCGENLCK', $db_adapter);

$rowset = $table_gateway->select();

$count = 0;
foreach ($rowset as $row) {
    $count++;
}

echo PHP_EOL . "found " . $count . " row(s)" . PHP_EOL;
```

result:
```
$ php pdo_test.php

Fatal error: Uncaught exception 'PDOException' with message 'SQLSTATE[42704]: Undefined object: -204 SCGENLCK in ETCWEBDB type *FILE not found. (SQLPrepare[-204] at /patched-php-src-5.6.10/php-5.6.10/ext/pdo_ibm/ibm_driver.c:145)' in /www/apps/EagleERP/vendor/zendframework/zend-db/src/Adapter/Driver/Pdo/Statement.php:186
Stack trace:
#0 /www/apps/EagleERP/vendor/zendframework/zend-db/src/Adapter/Driver/Pdo/Statement.php(186): PDO->prepare('SELECT SCGENLCK...')
#1 /www/apps/EagleERP/vendor/zendframework/zend-db/src/Adapter/Driver/Pdo/Statement.php(212): Zend\Db\Adapter\Driver\Pdo\Statement->prepare()
#2 /www/apps/EagleERP/vendor/zendframework/zend-db/src/TableGateway/AbstractTableGateway.php(239): Zend\Db\Adapter\Driver\Pdo\Statement->execute()
#3 /www/apps/EagleERP/vendor/zendframework/zend-db/src/TableGateway/AbstractTableGateway.php(209): Zend\Db\TableGateway\AbstractTableGateway->executeSelect(Object(Zend\Db\Sql\Select))
#4 /www/apps/EagleERP/vendor/zendframework/zend-db/src/TableGateway/AbstractTableGateway.php(196):  in /www/apps/EagleERP/vendor/zendframework/zend-db/src/Adapter/Driver/Pdo/Statement.php on line 186
```

ezimuel:
@alshayed this is a different issue, compared with the one that you submitted early. The key point can be summarized with your sentence:
> The problem is that PDO doesn't like receiving unsupported keys and ignores all other keys following them.

I will try to reproduce it and let you know. Thanks for your patience and the detailed explanation!

ezimuel:
Closing this PR due to unavailability of a IBM Db2 environment for testing.

PR #193, opened 2016-11-15
Create InsertIgnore

forw4rd:
i think thats missing :)


alextech:
I am thinking if this is too MySQL specific and too small of a change to warrant a new class. What will happen when I try to run this on PostgreSQL or any of the other adapters this library is supposed to support? For things like Sequence which is database specific, that goes into a `Feature` namespace, which augments standard SQL. Then, if a generic command has multiple ways of doing it across engines, there is a driver specific override. Currently, this solution falls into neither. I just have to know InsertIgnore is bad in anything other than MySQL. Which is a reasonable assumption - I would not think of it unless I knew the engine. However, this is still pretty generic, and modification is small enough to an already mostly existing feature that perhaps its more useful to make into an API property.

$insert = new Insert();
$insert->ignoreDuplicate(true);

Then, this can be done at any engine level without knowledge of exactly what needs to be done, descriptive, and easy to document. IgnoreDuplicate can adjust specification based on true/false. In MySQL it would be as easy str_replac-ing after INSERT (then I can come around to do other engines later), and unit test will be easy to add too.

PR #196, opened 2016-11-29
Fix $valueSet initial value in In predicate

nanawel:
It should be possible to construct a IN predicate with empty values set.
```php
$select->where('mycolumn IN ?', []);
```
It's not possible with the current code because of the `if ($valueSet)` in the constructor and the missing default value in the field definition above, which leads to the following error on line 122 when predicate is built:
```
Invalid argument supplied for foreach()
```

This commit should fix that.
Thanks in advance.


nanawel:
Fix updated with a better implementation that does not break unit tests.

ezimuel:
@nanawel why do you want this? Can you provide a real example? 
Moreover, remember to send always unit tests on each PR, thanks!

nanawel:
@ezimuel I need this because the content of the set can be dynamic and not known beforehand, and in that case I certainly do not want the Zend framework to throw an ugly error on the `foreach()` while the result can be valid SQL for certain platforms (SQLite for example allows an empty set).
I prefer to allow initializing the predicate with an empty set and let the database return an error if it's not allowed.

This PR was messy so I closed it. Please see #223 instead. Thanks

Issue #192, opened 2016-11-15
Zend\DB\SQL  order() only allow one space between field name and  DESC|ASC  


comeUpWithItLater:
`
$select = $this->tableGateway->getSql() ->select();
//... 
$select  ->order("date_added DESC");  // one space  between  "date_added" and  "DESC"
echo  $select->getSqlString($this->dbAdapter->getPlatform());   // SELECT ... ORDER BY date_added DESC

$select  ->order("date_added   DESC");  // more than 2 spaces  between  "date_added" and  "DESC"
echo  $select->getSqlString($this->dbAdapter->getPlatform());   // SELECT ... ORDER BY date_added ASC <- not DESC !!!!
`

so  only one space is allowed between field name and  DESC|ASC, I can't say that this is a bug/ issue

alextech:
Thank you for report!

ezimuel:
Fixed in https://github.com/zendframework/zend-db/pull/205

PR #205, opened 2017-01-05
#192 spaces in order function

alextech:
Fixing #192 processing of ORDER BY correctly splits by multiple spaces, but ignores the fact that extra spaces are still there.

### DIFF
--- a/src/Sql/Select.php
+++ b/src/Sql/Select.php
@@ -651,7 +651,7 @@ class Select extends AbstractPreparableSql
                     $v = self::ORDER_ASCENDING;
                 }
             }
-            if (strtoupper(trim($v)) == self::ORDER_DESCENDING) {
+            if (strcasecmp(trim($v), self::ORDER_DESCENDING) === 0) {
                 $orders[] = [$platform->quoteIdentifierInFragment($k), self::ORDER_DESCENDING];
             } else {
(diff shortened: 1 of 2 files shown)


src/Sql/Select.php
```diff
@@ -651,7 +651,7 @@ protected function processOrder(PlatformInterface $platform, DriverInterface $dr
                     $v = self::ORDER_ASCENDING;
                 }
             }
-            if (strtoupper($v) == self::ORDER_DESCENDING) {
+            if (strtoupper(trim($v)) == self::ORDER_DESCENDING) {
```
Ocramius: Is there any guarantee that `$v` is already uppercase?

test/Sql/SelectTest.php
```diff
@@ -400,6 +400,13 @@ public function testOrder()
         $this->assertEquals(['name ASC', 'age DESC'], $select->getRawState('order'));
 
         $select = new Select;
+        $select->order('name  DESC');
```
Ocramius: This should be in a new test, not added to the same test.

src/Sql/Select.php
alextech: No, this is directly from input $select->order('col_name desc') or 'col_name DESC'. That said, case should not matter here just to find if defaults should be applied, so I will switch to case insensitive comparison.

ezimuel:
@alextech thanks for your contribution!

PR #209, opened 2017-01-23
Updated docblock on join to show TableIdentifer is allowed

dstockto:
In addition to accepting strings and arrays, TableIdentifier is allowed and sometimes necessary. This updates the doc block for Select and Join to include it so Intellisense in IDEs will show this could work.

### DIFF
--- a/src/Sql/Join.php
+++ b/src/Sql/Join.php
@@ -108,7 +108,7 @@ class Join implements Iterator, Countable
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Ocramius:
As discussed on IRC, passing in an identifier object is probably the safest way to deal with this API.

Ocramius:
Merged manually via zendframework/zend-db@cd85134aa221df262026666e60cf4e22b1816cf6

PR #210, opened 2017-01-25
Fix join() docblock

autowp:


michalbundyra:
No longer needed. As I see on current master it has been already addressed.

PR #216, opened 2017-02-10
MySQL "ADD COLUMN ... AFTER ..." support

nanawel:
ZF1 supports MySQL syntax allowing to add a column AFTER another one in an ALTER TABLE statement. ZF2 does not. This PR is intended to fix that.


alextech:
I want to say "needs a test" but looking at the decorator, there are no existing tests for it where you could have added it. Yet it must be done :) Can you please create an `AlterTableDecoratorTest` in the same place and same way `CreateTableDecoratorTest` and add your option in there? (I could assist if unsure how)

nanawel:
Sure, I'll add it to the PR.

alextech:
Thank you :)

nanawel:
Yes, this is my assumption too.

nanawel:
New test class included, no error detected on PHPUnit on my instance.

ezimuel:
@nanawel thanks for your contribution!

PR #217, opened 2017-02-11
Changed copyright to 2017

chr-is:


ezimuel:
@chr-is as reported in #218 can you send this PR against master? Develop should be used only for new features. Thanks!

malukenho:
@ezimuel I guess it's invalid as per @weierophinney explanation in here https://github.com/zendframework/zend-expressive/pull/418#issuecomment-270484049

ezimuel:
@malukenho thanks, I didn't notice that. @chr-is we don't need to update the copyright date in all the files, only in `LICENSE.md`. Can you create a new PR with only this change? Thanks!

chr-is:
@ezimuel see #225

PR #220, opened 2017-02-14
Added SQL Server to Vagrant + fixed tests

ezimuel:
This PR proposes a new Vagrant box using Ubuntu 16.04 with MySQL, PostgreSQL and [SQL Server for Linux](https://www.microsoft.com/en-us/sql-server/sql-server-vnext-including-Linux). The Vagrant box uses a local IP address `192.168.20.20` to facilitate the unit tests for environments that already have one or more database installed. 

I also refactored also `phpunit.xml.dist` using the new ip address and the new default passwords. All the integration test are disabled by default. You need to enable it changing `false` to `true`. For instance, this is the value to change in `phpunit.xml.dist` to execute the integration test of SQL Server:
```xml
<env name="TESTS_ZEND_DB_ADAPTER_DRIVER_SQLSRV" value="true" />
```

### DIFF
--- a/.ci/mysql_fixtures.sh
+++ b/.ci/mysql_fixtures.sh
@@ -2,4 +2,4 @@
 
 echo "Configure MySQL test database"
 
-mysql -u root -pPassword123 -e 'create database zenddb_test;'
+mysql --user=root --password=Password123 -e 'create database zenddb_test;'
--- a/.ci/mysql_fixtures.sh
+++ b/.ci/mysql_fixtures.sh
@@ -2,4 +2,4 @@
 
 echo "Configure MySQL test database"
 
-mysql -u root -e 'create database zenddb_test;'
+mysql -u root -pPassword123 -e 'create database zenddb_test;'
--- a/.ci/mysql_fixtures_travis.sh
+++ b/.ci/mysql_fixtures_travis.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+echo "Configure MySQL test database"
+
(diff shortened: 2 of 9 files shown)


alextech:
SqlServer addition is very welcome. My favorite feature is superior management of prepared statements of any vendor so far. PostgreSQL supposed to have a hint of that in version 10.

IMHO, if we are adding new engines, they should be rightfully supported. As it is now, it is a bit embarrassing, how despite having decorators for other engines, they are inaccurate and break on first usage attempt.  Yes, they are not officially listed as supported in documentation, as was pointed out to me, but my code editor says they are! I submitted findings so far but non strict ANSI 92 engines are treated with lower priority.   

I am happy with this PR initiative, but would like to voice an opinion that it should not be merged in until these 2 engines are updated thoroughly in both DBAL and DDL, otherwise I can see it getting disrespect in public.

Finally, is there any roadmap of this package, in terms of priorities and general timing? Hard to make up contribution scheduling not knowing and can only do so based on assumptions. For example, you having added SqlServer at vagrant to me implies higher priority for DDL, since vagrant fixture can be a prime place to put migrations.  But it conflicts with feedback I got earlier that DDL packages are lower priority. Conflicts, because in my mind, at least, the act of automated database creations promotes the idea of automated schema creation along with it.

ezimuel:
@alextech this PR is actually less pretentious. It only adds the SQL Server Linux installation on Vagrant for improve the integration tests of zend-db. With a working Vagrant box we can improve the code coverage and improve also the adapters with new features like DDL.

PR #223, opened 2017-02-27
Add support for empty values set with IN predicate

nanawel:
(Original PR #196)

It should be possible to construct a IN predicate with empty values set.
```php
$select->where('mycolumn IN ?', []);
```
Of course this code will most likely never be written as is but the content of the set can instead be the result of a calculation.

It's not possible with the current code because of the `if ($valueSet)` in the constructor and the missing default value in the field definition above, which leads to the following error on line 122 when predicate is built:
```
Invalid argument supplied for foreach()
```
Empty sets are supported by some DBMS (SQLite for example) and in that case are the equivalent of `WHERE FALSE` (since no result can be part of an empty set...).

### DIFF
--- a/src/Sql/Predicate/In.php
+++ b/src/Sql/Predicate/In.php
@@ -33,7 +33,7 @@ class In extends AbstractExpression implements PredicateInterface
         if ($identifier) {
             $this->setIdentifier($identifier);
         }
-        if ($valueSet) {
+        if ($valueSet !== null) {
             $this->setValueSet($valueSet);
         }
     }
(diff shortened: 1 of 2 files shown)


src/Sql/Predicate/In.php
```diff
@@ -122,9 +122,10 @@ public function getExpressionData()
             foreach ($values as $argument) {
                 list($replacements[], $types[]) = $this->normalizeArgument($argument, self::TYPE_VALUE);
             }
+            $values = count($values) > 0 ? array_fill(0, count($values), '%s') : [];
```
ezimuel: Why this? `array_fill()` returns `[]` if `$values === []`. Please remove this change, the previous code is ok.

nanawel:
@ezimuel Not with versions of PHP < 5.6. My code is here to avoid a warning with those older versions of PHP.
Refer to History section at https://secure.php.net/manual/fr/function.array-fill.php

ezimuel:
@nanawel ok, I see the point. Makes no sense to add a check for PHP < 5.6. The code looks good to me. I'm going to merge in a while. Thanks for your contribution!

nanawel:
And thanks to you!

PR #224, opened 2017-03-03
Fixes the special characters in bindParam for PDO

ezimuel:
This PR solve #35 and replaces #178. The $name in bindParam are replaced with the following, due to the PDO restriction [0-9a-zA-Z_]:
```php
return ':' . md5($name);
```
For instance, a filed name `test$` will be translated in `:79f5d7a7e10367f279c2500dcd03b3de` as bind param name. I used the `md5()` approach after some benchmark tests, see comments and commits below. I also provided unit test and integrational test.

### DIFF
--- a/test/Adapter/Driver/Pdo/StatementTest.php
+++ b/test/Adapter/Driver/Pdo/StatementTest.php
@@ -129,7 +129,7 @@ class StatementTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
-     * @see https://github.com/zendframework/zend-db/pull/178
+     * @see https://github.com/zendframework/zend-db/pull/224
      */
     public function testExecuteWithSpecialCharInBindParam()
     {
@@ -137,7 +137,13 @@ class StatementTest extends \PHPUnit_Framework_TestCase
         $this->statement->setDriver(new Pdo(new Connection($testSqlite)));
         $this->statement->initialize($testSqlite);
 
-        $this->statement->prepare('INSERT INTO test (text_, text$) VALUES (:text_, :text_36_)');
+        $this->statement->prepare(
+            'INSERT INTO test (text_, text$) VALUES (:' .
+            md5('text_') .
+            ', :' .
+            md5('text$') .
+            ')'
+        );
         $result = $this->statement->execute([ 'text_' => 'foo', 'text$' => 'bar']);
         $this->assertInstanceOf(Result::class, $result);
         $this->assertTrue($result->valid());
--- a/test/Adapter/Driver/Pdo/StatementTest.php
+++ b/test/Adapter/Driver/Pdo/StatementTest.php
@@ -137,13 +137,11 @@ class StatementTest extends \PHPUnit_Framework_TestCase
         $this->statement->setDriver(new Pdo(new Connection($testSqlite)));
         $this->statement->initialize($testSqlite);
 
-        $this->statement->prepare(
-            'INSERT INTO test (text_, text$) VALUES (:' .
-            md5('text_') .
-            ', :' .
-            md5('text$') .
-            ')'
-        );
+        $this->statement->prepare(sprintf(
+            'INSERT INTO test (text_, text$) VALUES (:%s, :%s)',
+            md5('text_'),
+            md5('text$')
+        ));
         $result = $this->statement->execute([ 'text_' => 'foo', 'text$' => 'bar']);
         $this->assertInstanceOf(Result::class, $result);
         $this->assertTrue($result->valid());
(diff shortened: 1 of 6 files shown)


ezimuel:
In the last commit, I changed the usage of `preg_replace_callback` in favor of `md5()` for performance reason. I did some [benchmark test](https://gist.github.com/ezimuel/5d9320d4384a23a42a96b180db7747ab) showing a big difference. MD5 if from 3x to 4x faster than `preg_replace_callback`.

alshayed:
Hi, is there any expected time frame for this to get merged in and released? Thanks!

ezimuel:
@alshayed I'll release in two weeks, sorry for the delay.

test/Adapter/Driver/Pdo/StatementTest.php
```diff
@@ -126,4 +127,30 @@ public function testExecute()
...
+        $this->statement->setDriver(new Pdo(new Connection($testSqlite)));
+        $this->statement->initialize($testSqlite);
+
+        $this->statement->prepare(
+            'INSERT INTO test (text_, text$) VALUES (:' .
+            md5('text_') .
+            ', :' .
+            md5('text$') .
+            ')'
+        );
```
weierophinney: I'd use `sprintf` here to make it more readable:

```php
$this->statement->prepare(sprintf(
    'INSERT INTO test (text_, text$) VALUES (:%s, :%s)',
    md5('text_'),
    md5('text$')
));
```

test/Adapter/Driver/Pdo/TestAsset/SqliteMemoryPdo.php
```diff
@@ -13,8 +13,19 @@ class SqliteMemoryPdo extends \Pdo
...
+        // execute the sql statement if not empty
+        if (!empty($sql)) {
+            if (false === $this->exec($sql)) {
+                throw new \Exception(sprintf(
+                    "Error: %s, %s",
+                    $this->errorCode(),
+                    implode(",", $this->errorInfo())
+                ));
+            }
+        }
```
weierophinney: Return early if `$sql` is empty:

```php
if (empty($sql)) {
    return;
}

if (false === $this->exec($sql)) {
    throw new \Exception(/* ... */);
}
```

test/Adapter/Driver/Pdo/StatementTest.php
ezimuel: Changed.

ezimuel:
I'm removing this PR because introduces issue like #288. It's not possible to fix #35 because of PHP bug [43130](https://bugs.php.net/bug.php?id=43130). This bug won't fix because of [this](http://news.php.net/php.cvs/47302) and [this](http://news.php.net/php.cvs/46848).

webimpress:
@ezimuel 
> It's not possible to fix #35

It is possible, IMHO.
What if we also replace fields in the query, not only on binding?

ezimuel:
@webimpress it's not safe to use characters like `-` in bind param name. Please, read the php.cvs [46848](http://news.php.net/php.cvs/46848) and [47302](http://news.php.net/php.cvs/47302).

webimpress:
@ezimuel yes, I've seen them. And as I understand you replace names with md5 names, but you are not changing these names in the query string but only on binding. What if we change them in both places?

ezimuel:
@webimpress It's not considered a best practice to use these characters in bind param name. Why you want to allow a bad practice in our code base? Moreover, it will introduce high CPU usage, with preg_replace() + md5() execution for each param in SQL statements. This is not a reliable solution.

webimpress:
@ezimuel  Ok, fine then, so the #35 should be marked as "invalid" or "won't fix" and never tried to be fixed :smile:

ezimuel:
@webimpress I know, this was my fault, sorry.

webimpress:
@ezimuel It's fine, you've tried to fix #178. Everyone makes mistake...
Good that you handle it very quickly and we will have shortly 2.9.1. Thanks for it! :)

alextech:
Automatically changing anything in the query will make it very difficult to debug at database level. A common debugging procedure is looking through database logs, in SqlServer watching through real time activity monitor. Then, copy those queries and search for them in the codebase. Changing the names will no longer match logged queries with whats in code.

I think instead should do validation for special chars at the same place you tried md5() and throw `InvalidArgumentException` to give a nice message before letting it crash at driver level.

ezimuel:
@alextech I added the throw Exception for invalid characters `[^a-zA-Z0-9_]` in my revert PR #289.

PR #225, opened 2017-03-05
Changed copyright to 2017

chr-is:


ezimuel:
@chr-is thanks for your contribution!

PR #230, opened 2017-03-11
Documentation error for TableIdentifier usage as alias.

### LINKED ISSUE #159: How to alias table name in TableIdentifier (forces string)
Before commit https://github.com/zendframework/zend-db/commit/09c67e365177701487a45733a0223fd0c26d3ef3#diff-35f95381c3fd67a5e9b9014a74c266a8 it was allowed to pass array ['t' => 'table'] in constructor ([documentation](https://github.com/zendframework/zend-db/blob/master/doc/book/sql.md) show aliasing as `new TableIdentifier(['t' => 'table'])`). Now constructor allows only string or callable with __toString method.
I see it's very inconsistent. Zend\Db\Sql\Select allows string, array or TableIdentifier. Insert, Update and Delete null|string|TableIdentifier. 
TableGateway -> getTable may return string or TableIdentifier (getTable->getTable chain).
RowGateway is using string or TableIdentifier.

Maybe there should be another class for normalizing this? I'm using postgres with schemas so it's much more noticeable (only TableIdentifier is able to pass schema).

alextech:
Fix #159

### DIFF
--- a/doc/book/sql.md
+++ b/doc/book/sql.md
@@ -150,7 +150,7 @@ $select->from(['t' => 'table']);
 
 // Using a Sql\TableIdentifier:
 // (same output as above)
-$select->from(new TableIdentifier(['t' => 'table']));
+$select->from(['t' => new TableIdentifier('table')]);
 ```
 
 ### columns()


ezimuel:
@alextech thanks for your contribution!

Issue #161, opened 2016-09-06
SequenceFeature - invalid quoting for Postgresql adapter


xorock:
I'm using Postgresql schemas (without SET search_path). What I'm trying to achieve is getting getLastGeneratedValue with TableGateway when in transaction. I've added `new SequenceFeature('id', 'site.table_id_seq"')` but this throws PDOException. Generated query is
SELECT NEXTVAL('"site.table_id_seq"') and should be SELECT NEXTVAL('"site"."table_id_seq"').

To fix this You should allow arrays as second parameter and change

```php
case 'PostgreSQL':
                $sql = 'SELECT NEXTVAL(\'"' . $this->sequenceName . '"\')';
```

to

```php
case 'PostgreSQL':
                $sql = 'SELECT NEXTVAL(\'' . $platform->quoteIdentifierChain($this->sequenceName) . '\')';
```

michalbundyra:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/99.

PR #233, opened 2017-03-11
Delegate quoteIdentifier operation to quoteIdentifierChain

alextech:
Clean up identifier quoting.  quoteIdentifierChain was doing the same thing quoteIdentifier was doing but without safety of using predefined escape tokens. In turn, quoteIdentifier was producing same result as quoteIdentifierChain by adding quotes around the token, which can be simplified by immediately converting identifier to array, pass it to identiferChain, and achieve same result.  Overriding is now necessary only if there are really different rules by the engine. Eg. DB2 seems to be doing something very different.

This will help a lot with consistency in DB engines that have multiple identifiers in the path of a database objects. Eg. PostgreSQL always has trouble between choosing identifier vs chain. This can potentially solve a good portion of bad syntax generation for them.   Could consider deprecating one of them.

### DIFF
--- a/src/Adapter/Platform/AbstractPlatform.php
+++ b/src/Adapter/Platform/AbstractPlatform.php
@@ -1,9 +1,11 @@
 <?php
+
 /**
- * Zend Framework (http://framework.zend.com/)
+ * Zend Framework (http://framework.zend.com/).
  *
  * @link      http://github.com/zendframework/zf2 for the canonical source repository
- * @copyright Copyright (c) 2005-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ *
+ * @copyright Copyright (c) 2005-2017 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   http://framework.zend.com/license/new-bsd New BSD License
  */
 
@@ -31,7 +33,7 @@ abstract class AbstractPlatform implements PlatformInterface
      */
     public function quoteIdentifierInFragment($identifier, array $safeWords = [])
     {
-        if (! $this->quoteIdentifiers) {
+        if (!$this->quoteIdentifiers) {
             return $identifier;
         }
 
@@ -54,8 +56,8 @@ abstract class AbstractPlatform implements PlatformInterface
             $identifier .= isset($safeWordsInt[strtolower($part)])
                 ? $part
                 : $this->quoteIdentifier[0]
-                . str_replace($this->quoteIdentifier[0], $this->quoteIdentifierTo, $part)
-                . $this->quoteIdentifier[1];
+                .str_replace($this->quoteIdentifier[0], $this->quoteIdentifierTo, $part)
+                .$this->quoteIdentifier[1];
         }
 
         return $identifier;
@@ -66,13 +68,7 @@ abstract class AbstractPlatform implements PlatformInterface
      */
     public function quoteIdentifier($identifier)
     {
-        if (! $this->quoteIdentifiers) {
-            return $identifier;
-        }
-
-        return $this->quoteIdentifier[0]
-            . str_replace($this->quoteIdentifier[0], $this->quoteIdentifierTo, $identifier)
-            . $this->quoteIdentifier[1];
+        return $this->quoteIdentifierChain([$identifier]);
     }
 
     /**
@@ -80,7 +76,21 @@ abstract class AbstractPlatform implements PlatformInterface
      */
     public function quoteIdentifierChain($identifierChain)
     {
-        return '"' . implode('"."', (array) str_replace('"', '\\"', $identifierChain)) . '"';
+        if (is_string($identifierChain)) {
+            $identifierChain = [$identifierChain];
+        }
+
+        if (!$this->quoteIdentifiers) {
+            return implode($this->getIdentifierSeparator(), $identifierChain);
+        }
+
+        /** @var array $identifierChain */
+        foreach ($identifierChain as $key => $identifier) {
+            $identifierChain[$key] = str_replace($this->quoteIdentifier[0], $this->quoteIdentifierTo, $identifier);
+        }
+        $chainGlue = $this->quoteIdentifier[1].$this->getIdentifierSeparator().$this->quoteIdentifier[0];
+
+        return $this->quoteIdentifier[0].implode($chainGlue, $identifierChain).$this->quoteIdentifier[1];
     }
 
     /**
@@ -105,10 +115,11 @@ abstract class AbstractPlatform implements PlatformInterface
     public function quoteValue($value)
     {
         trigger_error(
-            'Attempting to quote a value in ' . get_class($this) .
+            'Attempting to quote a value in '.get_class($this).
             ' without extension/driver support can introduce security vulnerabilities in a production environment'
         );
-        return '\'' . addcslashes((string) $value, "\x00\n\r\\'\"\x1a") . '\'';
+
+        return '\''.addcslashes((string) $value, "\x00\n\r\\'\"\x1a").'\'';
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/Adapter/Platform/AbstractPlatform.php
```diff
@@ -1,9 +1,11 @@
 <?php
+
```
froschdesign: Please revert. The topic of this PR is "Delegate quoteIdentifier operation to quoteIdentifierChain" and not fixing the CS!

src/Adapter/Platform/AbstractPlatform.php
```diff
@@ -31,7 +33,7 @@
      */
     public function quoteIdentifierInFragment($identifier, array $safeWords = [])
     {
-        if (! $this->quoteIdentifiers) {
+        if (!$this->quoteIdentifiers) {
```
froschdesign: Please revert, the white space was correct.

src/Adapter/Platform/AbstractPlatform.php
```diff
@@ -54,8 +56,8 @@ public function quoteIdentifierInFragment($identifier, array $safeWords = [])
             $identifier .= isset($safeWordsInt[strtolower($part)])
                 ? $part
                 : $this->quoteIdentifier[0]
-                . str_replace($this->quoteIdentifier[0], $this->quoteIdentifierTo, $part)
-                . $this->quoteIdentifier[1];
+                .str_replace($this->quoteIdentifier[0], $this->quoteIdentifierTo, $part)
```
froschdesign: Please revert, the white space was correct.

src/Adapter/Platform/AbstractPlatform.php
```diff
@@ -105,18 +115,19 @@ public function getQuoteValueSymbol()
     public function quoteValue($value)
     {
         trigger_error(
-            'Attempting to quote a value in ' . get_class($this) .
+            'Attempting to quote a value in '.get_class($this).
```
froschdesign: Please revert, the white space was correct.

src/Adapter/Platform/AbstractPlatform.php
```diff
@@ -105,18 +115,19 @@ public function getQuoteValueSymbol()
...
             ' without extension/driver support can introduce security vulnerabilities in a production environment'
         );
-        return '\'' . addcslashes((string) $value, "\x00\n\r\\'\"\x1a") . '\'';
+
+        return '\''.addcslashes((string) $value, "\x00\n\r\\'\"\x1a").'\'';
```
froschdesign: Please revert, the white space was correct.

src/Adapter/Platform/AbstractPlatform.php
```diff
@@ -105,18 +115,19 @@ public function getQuoteValueSymbol()
...
     }
 
     /**
      * {@inheritDoc}
      */
     public function quoteTrustedValue($value)
     {
-        return '\'' . addcslashes((string) $value, "\x00\n\r\\'\"\x1a") . '\'';
+        return '\''.addcslashes((string) $value, "\x00\n\r\\'\"\x1a").'\'';
```
froschdesign: Please revert, the white space was correct.

src/Adapter/Platform/AbstractPlatform.php
```diff
@@ -66,21 +68,29 @@ public function quoteIdentifierInFragment($identifier, array $safeWords = [])
...
      * {@inheritDoc}
      */
     public function quoteIdentifierChain($identifierChain)
     {
-        return '"' . implode('"."', (array) str_replace('"', '\\"', $identifierChain)) . '"';
+        if (is_string($identifierChain)) {
+            $identifierChain = [$identifierChain];
+        }
+
+        if (!$this->quoteIdentifiers) {
```
froschdesign: White space after exclamation mark is missing.

src/Adapter/Platform/AbstractPlatform.php
```diff
@@ -66,21 +68,29 @@ public function quoteIdentifierInFragment($identifier, array $safeWords = [])
...
+            return implode($this->getIdentifierSeparator(), $identifierChain);
+        }
+
+        /** @var array $identifierChain */
+        foreach ($identifierChain as $key => $identifier) {
+            $identifierChain[$key] = str_replace($this->quoteIdentifier[0], $this->quoteIdentifierTo, $identifier);
+        }
+        $chainGlue = $this->quoteIdentifier[1].$this->getIdentifierSeparator().$this->quoteIdentifier[0];
```
froschdesign: White space before and after the dots is missing.

src/Adapter/Platform/AbstractPlatform.php
```diff
@@ -66,21 +68,29 @@ public function quoteIdentifierInFragment($identifier, array $safeWords = [])
...
+
+        return $this->quoteIdentifier[0].implode($chainGlue, $identifierChain).$this->quoteIdentifier[1];
```
froschdesign: White space before and after the dots is missing.

froschdesign:
@alextech 
> No unit tests added because this is well covered by existing cases; nothing new or different is generated.

Thanks for the hint!

src/Adapter/Platform/AbstractPlatform.php
```diff
@@ -115,19 +111,18 @@ public function getQuoteValueSymbol()
     public function quoteValue($value)
     {
         trigger_error(
-            'Attempting to quote a value in '.get_class($this).
+            'Attempting to quote a value in ' . get_class($this) .
```
froschdesign: Can we use `sprintf` here?

src/Adapter/Platform/AbstractPlatform.php
alextech: Are you reviewing old commits by accident? I think this is before you told me not to run outdated CS fixer in another PR. I redid all my other PRs same evening, including this one.

src/Adapter/Platform/AbstractPlatform.php
alextech: Yes we can. Was not sure if you wanted it saved in variable or not. Made it into parameter. And code standard for dot at beginning of line does not seem to apply here, since no equal sign so I put it at the end.

alextech:
How does that one small sprintf() change lowers unit test coverage across all files?

ezimuel:
@alextech you mentioned that this PR fixes some potential issue in SQL syntax generation. Can we found and example? We need to prove with a unit test that this PR fixes something. Now, we have the same unit test of before and this PR seems not necessary.

alextech:
@ezimuel pretty much every quoteIdentifier call across the codebase when using PostgreSQL or SQL Server **including userland usage in other libraries and application code** 

https://github.com/zendframework/zend-db/pull/247/files/bfb73c8bf4be5d98207a9911eca14aabdadd79c8#diff-0113773e7838ad267d6360c243edc298R235 
Or https://github.com/zendframework/zend-db/issues/161 (Ignore the sequence part)

Or related random example from other library https://github.com/zendframework/zend-log/issues/63,
some of my DDL PRs...

The reoccurring theme is the need to check database type and if an object path was given **for every** attempt to quote an identifier to decide if `quoteIdentifier` or `quoteIdentifierChain` should be used, both within zend db codebase, **and in application code** using zend db.  Any place where this is forgotten, such as pull request I linked, results in crashes that @xorock brought up in several places. As I was investigating why that is, I noticed how both functions essentially do the same thing with minor difference so I consolidated the usage. The **lack of unit tests in itself proves** that the PR does not break any existing usage; the whole point is to fix this inconsistency that makes *all existing* code work as it did before, while not needing to check database type.

For explanation of what I mean by "object path", look last comment in #206 and a typical usage example in the issue description. ( I cannot remember if this PR can fix that issue directly, but its a step towards it, as with it, identifiers can be expressed as an array or if string given, explode on object path delimiter to create needed array, and be able to quote it without having to worry about it being chain or what)

weierophinney:

michalbundyra:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/77.

PR #240, opened 2017-03-30
Updated the mssql configuration line to reflect the new mssql-conf

jschwartz73:
The step to configure MS SQL server has changed since this Vagrant file was created.  Updated the vagrant file with the fix.


ezimuel:
@jschwartz73 Thanks! I removed only `sudo` in the command, since all the `Vagrantfile` is executed as root.

PR #244, opened 2017-04-20
Add nested where clauses to documentation

### LINKED ISSUE #238: Expressions with 'or'
Not sure that is unexpected, but zf1 db works differently with 'or' in expressions

```php
$select = new Sql\Select();
$select->where(['column1 is null or column1 = 2']);
$select->where(['column2 = 3']);
        
print $select->getSqlString(); exit;
```

prints
```sql
SELECT * WHERE column1 is null or column1 = 2 AND column2 = 3
```

expected by me
```sql
SELECT * WHERE (column1 is null or column1 = 2) AND (column2 = 3)
```

xrmx:
Fix #238

### DIFF
--- a/doc/book/sql.md
+++ b/doc/book/sql.md
@@ -258,6 +258,23 @@ $select->from('foo')->where([
 ]);
 ```
 
+As another example of complex queries with nested conditions e.g.
+
+```sql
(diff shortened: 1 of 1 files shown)


ezimuel:
Thanks @xrmx for your contribution!

Issue #36, opened 2015-09-15
$AlterTable->dropConstraint() Throw error on MySQL,There is no support for "DROP CONSTRAINT"


solody:
I'm dropping a constraint on mysql5.6, code:

```php
$AlterTable = new Ddl\AlterTable($test_table_name);
$AlterTable->dropConstraint('text_UniqueKey');

echo $sql->buildSqlString($AlterTable,$adapter);

$adapter->query(
    $sql->buildSqlString($AlterTable,$adapter),
    $adapter::QUERY_MODE_EXECUTE
);
```
And this code output :
```sql
ALTER TABLE `test` DROP CONSTRAINT `text_UniqueKey`
```
Here is the problem, Mysql nerver support the Syntax "DROP CONSTRAINT", so it throw an error :
```bash
SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'CONSTRAINT `text_UniqueKey`' at line 2
```
Note: $adapter config array is 
```php
'db' => array(
    'driver' => 'Pdo_Mysql',
    'hostname' => 'localhost',
    'username' => 'root',
    'password' => 'abc123',
    'database' => 'easypay',
)
```
Help ! What to do now ???

nanawel:
Interested in this issue too. I guess there should be a way to do this through the MySQL `AlterTable` decorator.

alextech:
Yes, you can see example in my MSSQL PR at https://github.com/zendframework/zend-db/pull/231/files and more specifically https://github.com/zendframework/zend-db/pull/231/files#diff-1c04a65ef544d7ab52700dcd9818209e

Generally speaking:
1. Extend existing decorator into your project, and use that decorator (or modify sources in your own zend db fork)
2. Modify inherited $specifications property to provide a template for what syntax should be with `printf` style placeholders 
3. add `processWhatever` function, which matches the key of $specifications template you added, or trying to modify. It should return array that will be able to fill placeholders from step 2.

Did not realize some syntax does not work for mysql here. Usually its the other way around. What is appropriate mysql way of doing this?

nanawel:
> Did not realize some syntax does not work for mysql here. Usually its the other way around. What is appropriate mysql way of doing this?
In MySQL, `DROP CONSTRAINT` is replaced by `DROP FOREIGN KEY` for foreign key constraints only.

[...]

PR #247, opened 2017-05-22
Fix AlterTable::dropConstraint() by allowing the use of decorator

nanawel:
Fixes a broken implementation under MySQL.

See issue #36 

**Warning:** introduces a potential breaking change (string parameter is now a typed object).

### DIFF
--- a/src/Sql/Ddl/AlterTable.php
+++ b/src/Sql/Ddl/AlterTable.php
@@ -10,6 +10,7 @@
 namespace Zend\Db\Sql\Ddl;
 
 use Zend\Db\Adapter\Platform\PlatformInterface;
+use Zend\Db\Metadata\Object\ConstraintObject;
 use Zend\Db\Sql\AbstractSql;
 
 class AlterTable extends AbstractSql implements SqlInterface
@@ -27,7 +28,7 @@ class AlterTable extends AbstractSql implements SqlInterface
     protected $addColumns = [];
 
     /**
-     * @var array
+     * @var Constraint\ConstraintInterface[]
      */
     protected $addConstraints = [];
 
@@ -42,7 +43,7 @@ class AlterTable extends AbstractSql implements SqlInterface
     protected $dropColumns = [];
 
     /**
-     * @var array
+     * @var ConstraintObject[]
      */
     protected $dropConstraints = [];
 
@@ -74,7 +75,7 @@ class AlterTable extends AbstractSql implements SqlInterface
         ],
         self::DROP_CONSTRAINTS  => [
             "%1\$s" => [
-                [1 => "DROP CONSTRAINT %1\$s,\n", 'combinedby' => ""],
+                [2 => "DROP %1\$s %2\$s,\n", 'combinedby' => ""],
             ]
         ]
     ];
@@ -138,12 +139,12 @@ class AlterTable extends AbstractSql implements SqlInterface
     }
 
     /**
-     * @param  string $name
+     * @param  ConstraintObject $constraint
      * @return self Provides a fluent interface
      */
-    public function dropConstraint($name)
+    public function dropConstraint(ConstraintObject $constraint)
     {
-        $this->dropConstraints[] = $name;
+        $this->dropConstraints[] = $constraint;
 
         return $this;
     }
@@ -229,7 +230,10 @@ class AlterTable extends AbstractSql implements SqlInterface
     {
         $sqls = [];
         foreach ($this->dropConstraints as $constraint) {
-            $sqls[] = $adapterPlatform->quoteIdentifier($constraint);
+            $sqls[] = [
+                'CONSTRAINT',
+                $adapterPlatform->quoteIdentifier($constraint->getName())
+            ];
         }
 
         return [$sqls];
--- a/src/Sql/Platform/Mysql/Ddl/AlterTableDecorator.php
+++ b/src/Sql/Platform/Mysql/Ddl/AlterTableDecorator.php
@@ -248,4 +248,17 @@ class AlterTableDecorator extends AlterTable implements PlatformDecoratorInterfa
 
         return $columnA - $columnB;
     }
+
+    protected function processDropConstraints(PlatformInterface $adapterPlatform = null)
+    {
+        $sqls = [];
+        foreach ($this->dropConstraints as $constraint) {
+            $sqls[] = [
+                $constraint->getType(),
+                $adapterPlatform->quoteIdentifier($constraint->getName())
+            ];
+        }
+
+        return [$sqls];
+    }
 }
--- a/test/Sql/Ddl/AlterTableTest.php
+++ b/test/Sql/Ddl/AlterTableTest.php
@@ -9,6 +9,7 @@
 
 namespace ZendTest\Db\Sql\Ddl;
 
+use Zend\Db\Metadata\Object\ConstraintObject;
 use Zend\Db\Sql\Ddl\AlterTable;
 use Zend\Db\Sql\Ddl\Column;
 use Zend\Db\Sql\Ddl\Constraint;
@@ -66,8 +67,9 @@ class AlterTableTest extends \PHPUnit_Framework_TestCase
     public function testDropConstraint()
     {
         $at = new AlterTable();
-        $this->assertSame($at, $at->dropConstraint('foo'));
-        $this->assertEquals(['foo'], $at->getRawState($at::DROP_CONSTRAINTS));
+        $constraint = new ConstraintObject('foo', null);
+        $this->assertSame($at, $at->dropConstraint($constraint));
+        $this->assertEquals([$constraint], $at->getRawState($at::DROP_CONSTRAINTS));
     }
 
     /**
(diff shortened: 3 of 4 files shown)


src/Sql/Ddl/AlterTable.php
```diff
@@ -74,7 +75,7 @@ class AlterTable extends AbstractSql implements SqlInterface
         ],
         self::DROP_CONSTRAINTS  => [
             "%1\$s" => [
-                [1 => "DROP CONSTRAINT %1\$s,\n", 'combinedby' => ""],
+                [2 => "DROP %1\$s %2\$s,\n", 'combinedby' => ""],
```
alextech: Do not modify generic syntax from abstract classes other adapters inherit from for engine specific fixes. Fixing one engine breaks assumptions of others within this library or others making custom adapters.

src/Sql/Ddl/AlterTable.php
```diff
@@ -229,7 +230,10 @@ protected function processDropConstraints(PlatformInterface $adapterPlatform = n
     {
         $sqls = [];
         foreach ($this->dropConstraints as $constraint) {
-            $sqls[] = $adapterPlatform->quoteIdentifier($constraint);
+            $sqls[] = [
+                'CONSTRAINT',
+                $adapterPlatform->quoteIdentifier($constraint->getName())
```
alextech: Until this https://github.com/zendframework/zend-db/pull/233 goes through that fixes bugs in quoteIdentifier vs quoteIdentifierChain this will break PostgreSQL. If lucky and that PR gets merged prior to this, then its fine.

src/Sql/Platform/Mysql/Ddl/AlterTableDecorator.php
```diff
@@ -248,4 +248,17 @@ private function compareColumnOptions($columnA, $columnB)
...
         return $columnA - $columnB;
     }
+
+    protected function processDropConstraints(PlatformInterface $adapterPlatform = null)
+    {
+        $sqls = [];
+        foreach ($this->dropConstraints as $constraint) {
+            $sqls[] = [
+                $constraint->getType(),
+                $adapterPlatform->quoteIdentifier($constraint->getName())
```
alextech: As above

doc/book/sql-ddl.md
```diff
@@ -87,8 +87,21 @@ $table->changeColumn('name', Column\Varchar('new_name', 50));
...
+$constraint = new ConstraintObject('my_index', null);
+$table->dropConstraint($constraint);
+```
+
+Notice: On MySQL, you need to specify the type of constraint you want to drop:
+```php
+use \Zend\Db\Metadata\Object\ConstraintObject;
+
+$fkConstraint = new ConstraintObject('my_fk', null);
+$fkConstraint->setType('FOREIGN KEY');
```
alextech: Disagree. If you want your constraint to represent foreign key, then use ForeignKey class that already does this.

nanawel: Sure, I can do that. But as I said on issue #36, I thought it was better because it describes an existing constraint (metadata) that can be retrieved by analyzing the database, hence programatically.

alextech: Oh I just noticed how ConstraintObject lets you do that... Why such duality... :( I guess at the moment both would work, but to go forward initiative to migrate to PHP 7.1 and all related type inconsistency fixes as discussed on development channel, this lose handling needs to go away. Imagine others who will not notice this duplication of functionality in ConstraintObject and ForeignKey.

alextech: I see... In that case metadata analyzer is broken not assigning foreign key constraint to ForeignKey constraint object. I should not be penalized for wanting to use given by the framework ForeignKey object to describe foreign key constraint. I suppose that means both need to be handled by the Decorator at this point since that's what the library gives us. Or clean creation of ConstraintObject to create ForeignKey,  which would be further reaching change.

nanawel: > Or clean creation of ConstraintObject to create ForeignKey, which would be further reaching change.

`\Zend\Db\Metadata\Source\AbstractSource::getConstraint()` seems to be doing its job when analyzing the table and assigns the type of constraints correctly.

But yeah, when you dig that down you realize there are some nasty inconsistencies to take care of ahead of that bug.

Anyway, I'll take your remarks into account and I'll make some changes, hopefully for the best. But I'm not going to rewrite half of the framework either.

alextech: Was about to suggest to let  `dropConstraint` method take in `AbstractConstraint` so you have access to best of both `ConstraintObject` and `ForeignKey`, but what do you know, `ConstraintObject` does not inherit from `AbstractConstraint` as name would imply!
Just cannot win here without
>rewrite half of the framework either.
But that's too much to ask for at this time ...

Great job on moving specifications to decorators. Those are tough to do.

For version 3, which according to discussions all new versions want to target PHP 7.1 strict type system, it would be needed to rewrite to tighten up the types allowed per method to weed out those 
>nasty inconsistencies to take care of ahead of that bug.
I would have fun doing it, but a little discouraging based on rate of PR reviews.

nanawel:
Fixed errors after merge with master. Ready for integration in 3.0.0 if everything's ok for you.

alextech:
@nanawel I too have this issue. (dont want to put in PR ID here to confuse the issue tracker relevancy, search "coverage" in open PRs where I point it out). Was passing then suddenly dropped. There was an effort to upgrade travis config files across all of ZF repositories to make it more strict and  after it there was more failings here in unexpected commits. I think I see why mine failed (though its a bit stretched) where I formatted my error output and so travis wants me to tests the full formatting logic, which is just a printf() XD ....

According to the report there is unexercised "processDropConstraints" function. Line was added there but none of your tests seem to cover it. Even if the whole file has coverage by previous commits, that does not mean *this* particular commit is covered, therefore travis thinks it is new untested BC break (which makes sense). That is, travis checks that *this* set of commits is fully covering all of changes.

Do your *new* tests execute that line 234 or only old ones?

nanawel:
There are actually 2 tests related to `processDropConstraints()` which are the [drop of the FK](https://github.com/zendframework/zend-db/pull/247/commits/07419e86371d88e1b26e65e4391f339152f49d21#diff-b4a5d500d7e4e0f8683a61e1a4ac7a3eR53) and the [index](https://github.com/zendframework/zend-db/pull/247/commits/07419e86371d88e1b26e65e4391f339152f49d21#diff-b4a5d500d7e4e0f8683a61e1a4ac7a3eR55). I added those lines way before today's merge so nothing has changed there. In fact, the only changes required after the merge were code style related, to match [how the strings are split up on `develop`](https://github.com/zendframework/zend-db/pull/247/commits/07419e86371d88e1b26e65e4391f339152f49d21#diff-b4a5d500d7e4e0f8683a61e1a4ac7a3eL59), which is slightly different from how they were when I submitted the PR.

nanawel:
Hi, still looking forward to seeing this in the upcoming 3.0.0. Thanks!

nanawel:
Hi,
just bumping this PR, in case it got forgotten :)

bjgajjar:
@alexdenvir — Can you please provide an update on this? 

I have a CMS and I am using Zend-DB. Right now I am not able to use the drop constraint functionality. I can use this code directly in my fork repo, but before that, I want to know the reason behind not letting this PR merged.

Update on this PR will be very helpful. So can you please provide it ASAP?

alexdenvir:
> @alexdenvir — Can you please provide an update on this?

Sorry, but no. Maybe try pinging @alextech instead?

alexdenvir:
Same here, although never to zend-db specifically. I saw you had left a review and assumed @bjgajjar meant to ping you instead. My bad

webimpress:
I see it's marked as BC Break so planned for v3 release. I guess it will be much easier to get it in without BC Break (if possible?)

webimpress:
@alextech @nanawel @alexdenvir @bjgajjar
I think it is possible to change the code to keep BC.
We can accept both - `string` and `ConstraintInterface`, and deprecate on string.
If string provided we can create `ConstraintObject` internally so behaviour should really stay the same. Then for v3 we will accept only `ConstraintInterface`.

What do you think?

bjgajjar:
@alexdenvir 

> Maybe try pinging @alextech instead?

Yeah. Sorry for the inconvenience.

@webimpress 

> We can accept both - `string` and `ConstraintInterface`, and deprecate on the string.

Agree we can do that way also.

@nanawel 

It seems like there is some issue with this PR. I tried to use this code in my fork repo and use the `dropConstraint` for `Unique` constraint as well as `PrimaryKey`. But not able to drop it.
When I print the SQL query it is returning "" as a constraint. 

                        DROP CONSTRAINT ""


For your reference;
You have been added `getConstraintType` function in `src/Sql/Platform/Mysql/Ddl/AlterTableDecorator.php` and when I alter the table it is using the `processDropConstraints` function of `src/Sql/Ddl/AlterTable.php`

webimpress:
@bjgajjar would you be able provide PR based on that with changes I've described above, fixes of the issue you found, and tests to cover changes, please?

itsmerhp:
@nanawel @webimpress @bjgajjar @alextech @alexdenvir 

> We can accept both - string and ConstraintInterface, and deprecate on string.
> If string provided we can create ConstraintObject internally so behaviour should really stay the same.

As suggested by @webimpress, I think we can provide BC(accept both - string and ConstraintInterface) by below change :
```
public function dropConstraint($constraint) {
	$constraint = is_object($constraint) ? $constraint : new ConstraintObject($constraint, NULL);
	$this->dropConstraints[] = $constraint;
	return $this;
}
```
Please share your thoughts on it.

weierophinney:

michalbundyra:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/72.

PR #249, opened 2017-05-28
Allow column names to contain any character

metalinspired:
In regards to issues #8 and #208 I've created a regex that will allow user to use any character as column name.
It will also keep functionality of allowing user to enter 'column as alias' as column name.
I've made an assumption that no one in right mind will have spaces on beginning or end of column name so they are ignored by regex.

### DIFF
--- a/src/Sql/AbstractSql.php
+++ b/src/Sql/AbstractSql.php
@@ -14,6 +14,7 @@ use Zend\Db\Adapter\ParameterContainer;
 use Zend\Db\Adapter\Platform\PlatformInterface;
 use Zend\Db\Sql\Platform\PlatformDecoratorInterface;
 use Zend\Db\Adapter\Platform\Sql92 as DefaultAdapterPlatform;
+use Zend\Db\Exception\RuntimeException;
 
 abstract class AbstractSql implements SqlInterface
 {
--- a/src/Sql/AbstractSql.php
+++ b/src/Sql/AbstractSql.php
@@ -409,8 +409,15 @@ abstract class AbstractSql implements SqlInterface
 
         if ($isIdentifier) {
             $matches = [];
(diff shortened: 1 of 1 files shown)


ezimuel:
@metalinspired  unit tests are missing for this PR, can you add it? Testing is very important here because you proposed a regex and we need to validate all the possible cases (or at least the most common). Thanks!

metalinspired:
@ezimuel I wouldn't know where to begin. The regex if written so that all existing tests are passing and as you can see I failed three times before succeeding. If the regex fails it will throw an exception and worse that can happen is malformed SQL syntax which was original problem to begin with. Personally I think that some of the undocumented behavior for column names, which I mentioned in #208, should be removed thus eliminating need for this patch.

ezimuel:
@metalinspired I'll try to work on this PR adding the unit test, thanks anyway for your contribution!

metalinspired:
@ezimuel Glad I could help

bjgajjar:
Hey @froschdesign @ezimuel - May we have any update for this? We need this in one of our products. So it'll be a great help if this PR gets merged.

webimpress:
@bjgajjar Still we need unit test for these changes. If you can provide these it will speed up the release. Thanks.

bjgajjar:
@metalinspired - It seems like there is some issue here.

- When tried to add a field named with `@!#!12123123`; DB will have a column with the name of ``@`!`#`!`12123123`` instead of `@!#!12123123`
- When adding any field with special character[`@user_id`] - Although it's getting stored in DB - It is throwing Unable to find field `@user_id` error with status code `404`.

weierophinney:

michalbundyra:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/70.

PR #252, opened 2017-06-09
fix Adapter\Driver\Pdo\Result::valid() problem

ares333:
problem:
When result count is 0 and no query method was executed(eg:current(),next()) $this->currentData is null type.So valid() will return true.
```
while ($list->valid()) {
         $list->current();//false if $list->count()===0
         $list->next();
}
```

### DIFF
--- a/src/Adapter/Driver/Pdo/Result.php
+++ b/src/Adapter/Driver/Pdo/Result.php
@@ -195,7 +195,7 @@ class Result implements Iterator, ResultInterface
      */
     public function valid()
     {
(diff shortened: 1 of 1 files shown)


Ocramius:
@ares333 this needs a test case

ares333:
yes you are right.test case is needed.
finally i find it should not be a bug.
A count() before iterating solve my problem.
if ($list->count() > 0) {
        while ($list->valid()) {
              $list->current();
              $list->next();
        }
}

PR #255, opened 2017-07-05
Bugfix/array should be valid datasource for resultset

bartmcleod:
When I upgraded from an older version of Zend\Db I found that certain unit tests in the client project were failing. When a datasource provided to a ResultSet is an array, the ResultSet should work out of the box. It didn't, because current() did not account for a datasource which has entered the system as just an array, while the buffer is flagged -1, when the ResultSet was initialized with such a datasource. So that flag on the buffer can be used to early out current() in such a situation.


waltertamboer:
If I'm not mistaken you've created the ReturnType class to force a null to be returned from `exchangeArray`. You might want to add that in the class' docblock to make that clear to other developers. Also try to remove the redundant new lines in that file to not make it messy (that's also why the build fails) :)

Other than that I don't see any problems.

test/ResultSet/ReturnType.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+
```
waltertamboer: You should remove the extra new lines, they make the build fail but it's also messy

test/ResultSet/ReturnType.php
```diff
@@ -0,0 +1,13 @@
...
+namespace ZendTest\Db\ResultSet;
+
+
+class ReturnType
```
waltertamboer: If I'm not mistaken you've created the ReturnType class to force a null to be returned from exchangeArray. You might want to add that in the class' docblock to make that clear to other developers.

bartmcleod:
Thanks @waltertamboer for the prompt review. I have update the ReturnType test class.

test/ResultSet/ReturnType.php
```diff
@@ -0,0 +1,16 @@
+<?php
+namespace ZendTest\Db\ResultSet;
+
+/**
+ * This class is just here to be able to provide the simplest possible custom return type
+ * for a ResultSet when calling ResultSet::setArrayObjectPrototype or using the constructor argument.
+ *
+ * Class ReturnType
```
waltertamboer: This is quite obvious ;-)

@weierophinney Not sure about the package tag? Does ZF have a standard for this?

test/ResultSet/ResultSetIntegrationTest.php
```diff
@@ -90,6 +90,20 @@ public function testCanProvideIteratorAsDataSource()
...
 
+    public function testCanProvideArrayAsDataSource()
+    {
+        $dataSource = [['foo']];
+        $this->resultSet->initialize($dataSource);
+        $this->assertEquals($dataSource[0], (array) $this->resultSet->current());
+
+        $returnType = new ReturnType();
+        $dataSource = [$returnType];
+        $this->resultSet->setArrayObjectPrototype($returnType);
```
samsonasik: I think it can use : 

```php
$dataSource = [new ArrayObject([], ArrayObject::ARRAY_AS_PROPS)];
$this->resultSet->setArrayObjectPrototype($dataSource);
```
so no need to have `ReturnType.php` file.

waltertamboer: Will that return null when exchangeArray is called?

bartmcleod: Thanks for your feedback. I don't think your suggestion covers the use case, while using ArrayObject obfuscates what I'm trying to prove, as ArrayObject is used all over ResultSet and it seems to know very well how to handle it. Problems arise when you are not using an ArrayObject, so I would rather not venture into using it. Also, when I try it, the test case `testCanProvideArrayAsDataSource` fails with `Object must be of type ArrayObject, or at least implement exchangeArray`.

bartmcleod: @waltertamboer It will return null.

samsonasik: @bartmcleod the object required for it is an object that need to has `exhangeArray()`, while `ArrayObject` already has it. I tried with ArrayObject and working fine at mine: 
screen shot 2017-07-05 at 8 02 14 pm

samsonasik: oops, sorry, wrong variable, will retry

samsonasik: should be ok with: 

```php
        $returnType = new ArrayObject([], ArrayObject::ARRAY_AS_PROPS);
        $dataSource = [$returnType];
        $this->resultSet->setArrayObjectPrototype($returnType);
        $this->resultSet->initialize($dataSource);

```

samsonasik: @bartmcleod I created PR to your PR branch https://github.com/bartmcleod/zend-db/pull/1

bartmcleod:
@samsonasik thanks for your contribution

ezimuel:
@bartmcleod if this is a bugfix why did you send to develop branch?

bartmcleod:
@ezimuel I haven't contributed in a while. Where should it go? master? Happy to change the target.

ezimuel:
@bartmcleod I can do that, no problem. I was just checking if this was a bugfix. Thanks!

bartmcleod:
@ezimuel Ok, thanks. It's a bugfix.

ezimuel:
@bartmcleod I tried to merge in master but it seems there are some issues. Can you try to send the PR against master? Thanks.

bartmcleod:
@ezimuel I will set a reminder for myself

ezimuel:
@bartmcleod I finally rebased the PR and merged in master (and develop). I'm closing the PR. Thanks again for your contribution!

bartmcleod:
@ezimuel welcome, and thanks for the merge. I had the reminder set for tomorrow morning, can clear that now :-)

PR #256, opened 2017-07-10
add Zend\Db\Sql\Predicate\Expression to docblocks at Predicate for $identifier parameter

samsonasik:


ezimuel:
@samsonasik thanks for your contribution!

PR #261, opened 2017-07-17
Fix Exception Firebird driver doesn't support lastInsertId

cgrandval:
When trying to use PDO with firebase, the following exception is raised: SQLSTATE[IM001]: Driver does not support this function: driver does not support lastInsertId()

The Firebird driver doesn't support the lastInsertId function. This function is called by getLastGeneratedValue function in every insert, update and delete and throw an exception.

This fix removes the call to lastInsertId when using firebase, based on PDO’s driverName.


ezimuel:
@cgrandval Thanks for your contribution!

Issue #267, opened 2017-09-26
Escaping of table, column and index name is wrong


rarog:
It seems that escaping is wrong for at least DDLs, I didn't test SELECT and other commands.
It goes wrong with (MySQL) and without (SQLite) decorators. The example for SQLite is flawed in terms of syntax, as the basic index creation isn't supported within CREATE TABLE command, but this should be a separate task similar to #163. This bug is only for the escaping issue.
Instead of just doubling the escape character there seems to be done some wrong escaping which is additionally run multiple times.

I tried following:
```php
$table = new CreateTable('t\'e"s`t');
$table->addColumn(new BigInteger('t\'e"s`tCol'))
    ->addColumn(new BigInteger('t\'e"s`tCol2'))
    ->addConstraint(new Constraint\PrimaryKey('t\'e"s`tCol'));
    ->addConstraint(new Index('t\'e"s`tCol2', 't\'e"s`tIndex'));
```

Expected output for MySQL:
```sql
CREATE TABLE `t'e"s``t` ( 
    `t'e"s``tCol` BIGINT NOT NULL AUTO_INCREMENT,
    `t'e"s``tCol2` BIGINT NOT NULL,
    PRIMARY KEY (`t'e"s``tCol`),
    INDEX `t'e"s``tIndex`(`t'e"s``tCol2`)
)
```
Actual output for MySQL:
```sql
CREATE TABLE `t'e"s``t` ( 
    `t``'``e``"``s``````tCol` BIGINT NOT NULL AUTO_INCREMENT,
    `t``'``e``"``s``````tCol2` BIGINT NOT NULL,
    PRIMARY KEY (`t``'``e``"``s``````tCol`),
    INDEX `t``'``e``"``s``````tIndex`(`t``'``e``"``s``````tCol2`)
)
```

Expected output for SQLite:
```sql
CREATE TABLE "t'e""s`t" ( 
    "t'e""s`tCol" BIGINT NOT NULL,
    "t'e""s`tCol2" BIGINT NOT NULL,
    PRIMARY KEY ("t'e""s`tCol"),
    INDEX "t'e""s`tCol2")
)
```
Actual output for SQLite:
```sql
CREATE TABLE "t'e's`t" ( 
    "t""'""e""'""s""`""tCol" BIGINT NOT NULL,
    "t""'""e""'""s""`""tCol2" BIGINT NOT NULL,
    PRIMARY KEY ("t""'""e""'""s""`""tCol"),
[...]

PR #268, opened 2017-09-27
Adding a fix for #267.

rarog:
In #267 I described the incorrect behaviour, when generating DDLs with characters in name that are used for escaping on different plattforms. This is an attempt to fix it.
I tested the behaviour and it corrects the script generation with the base (PostgreSQL, SQLite) and MySQL escaping behaviour.

MSSQL seems to behave like standard despite having separate escaping, so I didn't touch it. I was able to test with driver `Pdo_SqlServer`, perhaps it uses the escaping with brackets, if `Sqlsrv` driver is used, but I couldn't test it, because I don't have the corresponding extension installed. I get the RuntimeException with message `The Sqlsrv extension is required for this adapter but the extension is not loaded`. @alextech Can you test the example in #267 and add the [] brackets to every named element? If it uses the brackets for escaping and is wrong, is it fixed, if you change in `Zend\Db\Adapter\Platform\SqlSserver` $quoteIdentifierTo to
```php
    protected $quoteIdentifierTo = '[]';
```
At least from my tests SQL Management Studio escapes the opening bracket in names by adding the closing bracket to it, when I generate a CREATE TABLE script.

### DIFF
--- a/src/Adapter/Platform/AbstractPlatform.php
+++ b/src/Adapter/Platform/AbstractPlatform.php
@@ -19,7 +19,7 @@ abstract class AbstractPlatform implements PlatformInterface
     /**
      * @var string
      */
-    protected $quoteIdentifierTo = '\'';
+    protected $quoteIdentifierTo = '""';
 
     /**
      * @var bool
--- a/src/Adapter/Platform/Postgresql.php
+++ b/src/Adapter/Platform/Postgresql.php
@@ -17,13 +17,6 @@ use Zend\Db\Adapter\Exception;
 class Postgresql extends AbstractPlatform
 {
     /**
-     * Overrides value from AbstractPlatform to use proper escaping for Postgres
-     *
-     * @var string
-     */
-    protected $quoteIdentifierTo = '""';
-
-    /**
      * @var resource|\PDO
      */
     protected $resource = null;
--- a/src/Adapter/Platform/Sqlite.php
+++ b/src/Adapter/Platform/Sqlite.php
@@ -16,16 +16,6 @@ use Zend\Db\Adapter\Exception;
 class Sqlite extends AbstractPlatform
 {
     /**
-     * {@inheritDoc}
-     */
-    protected $quoteIdentifier = ['"','"'];
-
-    /**
-     * {@inheritDoc}
-     */
-    protected $quoteIdentifierTo = '\'';
-
-    /**
      * @var \PDO
      */
     protected $resource = null;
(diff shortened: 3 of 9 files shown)


src/Adapter/Platform/AbstractPlatform.php
```diff
@@ -19,7 +19,7 @@
     /**
      * @var string
      */
-    protected $quoteIdentifierTo = '\'';
+    protected $quoteIdentifierTo = '""';
```
mwillbanks: Why would you change the default, this is absolutely a MASSIVE BC BREAK.  I advise you to please no do that especially considering this could have been extended.

rarog: I could leave it as it is, but in that case I'd comment it, that it's only for BC reasons and contradicts the ANSI SQL standard.
I can't name any SQL implementation that uses single quotes for escaping if identifiers. The single quote is for strings.

I attach a wall of text aka TL;DR:

In chapter `5.2 <token> and <separator>` of 
[Information Technology - Database Language SQL](http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt) it's defined as

```
         <delimited identifier> ::=
              <double quote> <delimited identifier body> <double quote>

         <delimited identifier body> ::= <delimited identifier part>...

         <delimited identifier part> ::=
                <nondoublequote character>
              | <doublequote symbol>

         <nondoublequote character> ::= !! See the Syntax Rules

         <doublequote symbol> ::= <double quote><double quote>

         <delimiter token> ::=
                <character string literal>
              | <date string>
              | <time string>
              | <timestamp string>
              | <interval string>
              | <delimited identifier>
```

and same is recommended by [Apache docs](https://db.apache.org/derby/docs/10.1/ref/crefsqlj1003454.html):
```
Ordinary identifiers are identifiers not surrounded by double quotation marks. Delimited identifiers are identifiers surrounded by double quotation marks.
```

By standard DB/2, Oracle, PostgreSQL and SQLite use double quote for identifier escaping. Sybase and MS SQLServer default to quoting via [] but for example [SQLServer does support the double quote](https://docs.microsoft.com/en-us/sql/t-sql/statements/set-quoted-identifier-transact-sql) when `SET QUOTED_IDENTIFIER` is `ON`, and it is on, in ANSI mode:
```
When SET ANSI_DEFAULTS is ON, SET QUOTED_IDENTIFIER is enabled.
```

MySQL/MariaDB understand by default only the backtick, but also do support double quote in [ANSI mode](https://dev.mysql.com/doc/refman/5.5/en/compatibility.html):
`SET sql_mode = 'ANSI';` enables it for the running session.

rarog:
I'll work on unit test next week. Thx for the first review!

rarog:
@mwillbanks I finally found time to add the unit tests. Are they ok in that form?

weierophinney:

michalbundyra:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/65.

PR #271, opened 2017-10-07
Support dash on MySQL identifier

WellingGuzman:
On MySQL trying to select a column/table with dash in its name result in a error.

Ref #208


mwillbanks:
Dashes are horribly bad practice in identifiers for mysql, however, it is documented explicitly in https://dev.mysql.com/doc/refman/5.7/en/identifier-qualifiers.html therefore I can approve this PR.

ezimuel:
@WellingGuzman I merged this in `develop` since it's a new feature. Thanks for your contribution!

PR #259, opened 2017-07-13
Update Travis CI config

webimpress:
- removes hhvm build
- adds php 7.1, 7.2 ~~and nightly (7.3-dev) builds~~ (removed as not even alpha has been published)
- removes irc notifications
- adds composer.lock to the repository and run tests with lowest/locked/latest dependencies

PR #273, opened 2017-10-13
Implement our own error handler for db2_prepare

akrabat:
db2_prepare() may issue a [warning][1] in addition to returning false.
If you have an error handler set (e.g. in the Expressive pipeline) then
the message created is "db2_prepare(): Statement Prepare Failed" which
is not helpful.

Therefore, we create our own error handler and trap this ourselves so
that we can throw an ErrorException with more information from
db2_stmt_errormsg().

I've left the previous handling as I don't know PHP extension code well enough to be 100% sure that db2_prepare() always generates a warning when it returns false.

[1]: https://github.com/php/pecl-database-ibm_db2/blob/df913feb4dff56366ac4656b0bb6b39200794bde/ibm_db2.c#L4400

### DIFF
--- a/src/Adapter/Driver/IbmDb2/Statement.php
+++ b/src/Adapter/Driver/IbmDb2/Statement.php
@@ -9,6 +9,8 @@
 
 namespace Zend\Db\Adapter\Driver\IbmDb2;
 
+use ErrorException;
+use Throwable;
 use Zend\Db\Adapter\Driver\StatementInterface;
 use Zend\Db\Adapter\Exception;
 use Zend\Db\Adapter\ParameterContainer;
@@ -172,7 +174,14 @@ class Statement implements StatementInterface, Profiler\ProfilerAwareInterface
             $sql = $this->sql;
         }
 
-        $this->resource = db2_prepare($this->db2, $sql);
+        try {
+            set_error_handler($this->createErrorHandler());
+            $this->resource = db2_prepare($this->db2, $sql);
+        } catch (Throwable $e) {
+            throw new Exception\RuntimeException($e->getMessage() . '. ' . db2_stmt_errormsg(), db2_stmt_error(), $e);
+        } finally {
+            restore_error_handler();
+        }
 
         if ($this->resource === false) {
             throw new Exception\RuntimeException(db2_stmt_errormsg(), db2_stmt_error());
@@ -237,5 +246,32 @@ class Statement implements StatementInterface, Profiler\ProfilerAwareInterface
 
         $result = $this->driver->createResult($this->resource);
         return $result;
+    }
+
+    /**
+     * Creates and returns a callable error handler that raises exceptions.
+     *
+     * Only raises exceptions for errors that are within the error_reporting mask.
+     *
+     * @return callable
+     */
+    private function createErrorHandler() : callable
+    {
+        /**
+         * @param int $errno
+         * @param string $errstr
+         * @param string $errfile
+         * @param int $errline
+         * @return void
+         * @throws ErrorException if error is not within the error_reporting mask.
+         */
+        return function (int $errno, string $errstr, string $errfile, int $errline) : void {
+            if (! (error_reporting() & $errno)) {
+                // error_reporting does not include this error
+                return;
+            }
+
+            throw new ErrorException($errstr, 0, $errno, $errfile, $errline);
+        };
     }
 }
--- a/src/Adapter/Driver/IbmDb2/Statement.php
+++ b/src/Adapter/Driver/IbmDb2/Statement.php
@@ -177,7 +177,7 @@ class Statement implements StatementInterface, Profiler\ProfilerAwareInterface
         try {
             set_error_handler($this->createErrorHandler());
             $this->resource = db2_prepare($this->db2, $sql);
-        } catch (Throwable $e) {
+        } catch (ErrorException $e) {
             throw new Exception\RuntimeException($e->getMessage() . '. ' . db2_stmt_errormsg(), db2_stmt_error(), $e);
         } finally {
             restore_error_handler();
--- a/src/Adapter/Driver/IbmDb2/Statement.php
+++ b/src/Adapter/Driver/IbmDb2/Statement.php
@@ -10,7 +10,6 @@
 namespace Zend\Db\Adapter\Driver\IbmDb2;
 
 use ErrorException;
-use Throwable;
 use Zend\Db\Adapter\Driver\StatementInterface;
 use Zend\Db\Adapter\Exception;
 use Zend\Db\Adapter\ParameterContainer;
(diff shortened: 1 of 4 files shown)


Ocramius:
@akrabat you know exactly what I'm gonna ask you :-P

akrabat:
I can't run the integration tests against this client's IBM i DB2 and don't have my own, so unclear how to run them…

Ocramius:
Yeah, I know the scenario very well, but we're also not merging anything without verification, sorry...

akrabat:
Travis error is: 

```
PHP Parse error:  syntax error, unexpected ':', expecting ';' or '{' in /home/travis/build/zendframework/zend-db/vendor/doctrine/instantiator/src/Doctrine/Instantiator/Instantiator.php on line 95
```

Thoughts on how to fix?

Ocramius:
@akrabat I think the wrong `composer.lock` was committed to this repository

webimpress:
@Ocramius @akrabat tests on travis should works despite of committed composer.lock, maybe https://github.com/zendframework/zend-db/pull/259 will help

BTW currently we don't have composer.lock in zend-db at all

Ocramius:
Hmm, not sure how [this was installed](https://github.com/doctrine/instantiator/blob/185b8868aa9bf7159f5f953ed5afb2d7fcdc3bda/composer.json#L19) during the [composer update step](https://travis-ci.org/zendframework/zend-db/jobs/287492129#L596)

webimpress:
@Ocramius 
`composer install --no-interaction --ignore-platform-reqs`

https://travis-ci.org/zendframework/zend-db/jobs/287492129#L568
https://github.com/zendframework/zend-db/blob/master/.travis.yml#L81

`--ignore-platform-reqs` !!!

webimpress:
@Ocramius and maybe composer.lock is cached on travis?

src/Adapter/Driver/IbmDb2/Statement.php
```diff
@@ -238,4 +247,31 @@ public function execute($parameters = null)
...
     }
+
+    /**
+     * Creates and returns a callable error handler that raises exceptions.
+     *
+     * Only raises exceptions for errors that are within the error_reporting mask.
+     *
+     * @return callable
+     */
+    private function createErrorHandler() : callable
```
samsonasik: zend-db still use [php 5.5/5.6](https://github.com/zendframework/zend-db/blob/8dfeabc664dd45c5f1753fbb508db70d0e605bd4/composer.json#L16) so the return type won't work

src/Adapter/Driver/IbmDb2/Statement.php
```diff
@@ -238,4 +247,31 @@ public function execute($parameters = null)
...
+    {
+        /**
+         * @param int $errno
+         * @param string $errstr
+         * @param string $errfile
+         * @param int $errline
+         * @return void
+         * @throws ErrorException if error is not within the error_reporting mask.
+         */
+        return function (int $errno, string $errstr, string $errfile, int $errline) : void {
```
samsonasik: remove return type as well

src/Adapter/Driver/IbmDb2/Statement.php
```diff
@@ -172,7 +174,14 @@ public function prepare($sql = null)
             $sql = $this->sql;
         }
 
-        $this->resource = db2_prepare($this->db2, $sql);
+        try {
+            set_error_handler($this->createErrorHandler());
+            $this->resource = db2_prepare($this->db2, $sql);
+        } catch (Throwable $e) {
```
samsonasik: should catch `Exception` for php 5.5/5.6 fallback as well ?

src/Adapter/Driver/IbmDb2/Statement.php
samsonasik: also, `int` and `string` not supported in php 5.5/5.6

src/Adapter/Driver/IbmDb2/Statement.php
akrabat: No - in 5.6, nothing is thrown.

akrabat:
@Ocramius: Added an integration test.  I can't prove it works on every IBM DB2 installation out there, but it does on one specific IBMI I:

```
$ vendor/phpunit/phpunit/phpunit test/Adapter/Driver/IbmDb2/StatementIntegrationTest.php
PHPUnit 4.8.36 by Sebastian Bergmann and contributors.

.....

Time: 1.58 seconds, Memory: 4.00MB

OK (5 tests, 8 assertions)
```

akrabat:
@samsonasik Addressed your comments.

test/Adapter/Driver/IbmDb2/StatementIntegrationTest.php
```diff
@@ -86,6 +87,23 @@ public function testPrepare()
     }
 
     /**
+     * @covers Zend\Db\Adapter\Driver\IbmDb2\Statement::prepare
+     * @expectedException RuntimeException
```
webimpress: I would move it to function use, because right now it's not clear, which line should thrown that exception.

akrabat: @webimpress I can't - it doesn't work on the AS/400 when I do that. I have no idea why.

webimpress: Hm... so maybe exception is thrown not where you expecting it?

akrabat: It was a PHP error. I'll have another look when I'm next connected to that computer.

akrabat:
@webimpress Dunno what I did wrong last time, but it works this time!

```
$ vendor/phpunit/phpunit/phpunit  test/Adapter/Driver/IbmDb2/StatementIntegrationTest.php
PHPUnit 4.8.36 by Sebastian Bergmann and contributors.

.....

Time: 1.71 seconds, Memory: 4.00MB
```

test/Adapter/Driver/IbmDb2/StatementIntegrationTest.php
```diff
@@ -86,6 +87,23 @@ public function testPrepare()
...
+        $db2Resource = db2_connect(
+            $this->variables['database'],
+            $this->variables['username'],
+            $this->variables['password']
+        );
+        $statement = new Statement;
+        $statement->initialize($db2Resource);
+        $this->setExpectedException(RuntimeException::class);
+        $statement->prepare("SELECT");
+        unset($resource, $db2Resource);
```
webimpress: I think this line will be not even hit.

akrabat: It will if the test fails :)

mwillbanks:
@akrabat can you see if you can solve the coveralls although?

akrabat:
> can you see if you can solve the coveralls although?

@mwillbanks The coveralls problem is that I've added new code that is covered by an integration test rather than a unit test.

akrabat:
@mwillbanks Thought so more and then wrote some unit tests.

ezimuel:
@akrabat thanks for your contribution!

PR #276, opened 2017-10-31
Fix for PHP 7.2

remicollet:
From pr #260 without BC break


remicollet:
This fix both zend-db and zend-paginator test suite (in Fedora QA, with PHP 7.2.0RC5)

* https://apps.fedoraproject.org/koschei/package/php-zendframework-zend-paginator
* https://apps.fedoraproject.org/koschei/package/php-zendframework-zend-db

ezimuel:
@remicollet thanks!

PR #278, opened 2017-11-05
Drop PHP 5.5 Support

zf2timo:
This library had still support for PHP 5.5. This PR removes the PHP 5.5 version from Travis Configuration and also requires PHP 5.6 as minimum version for composer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,17 +50,8 @@ matrix:
     - php: 7.2
       env:
         - DEPS=latest
-    - php: 7.3
-      env:
-        - DEPS=lowest
-    - php: 7.3
-      env:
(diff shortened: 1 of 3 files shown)


ezimuel:
@zf2timo this PR contains a lot of changes and some rollback, can you open a new PR removing only the support of PHP 5.5 in travis and composer? Thanks!

webimpress:
@ezimuel it's not BC Break, so the label can be removed here

ezimuel:
@zf2timo closing this in favor of #282. @webimpress is fixing some Travis issue and removing the support of PHP 5.5. 
Thanks anyway for your contribution!

PR #280, opened 2017-11-10
pdo_dblib added version dsn parameter

vaclavvanik:
Currently it is not possible to set **version** parameter when connecting to MSSQL server via [pdo_dblib](http://php.net/manual/en/ref.pdo-dblib.connection.php#112146). This PR adds this parameter.


ezimuel:
@vaclavvanik thanks for your contribution!

PR #282, opened 2017-11-23
Hotfix - Travis CI Configuration

webimpress:
- removed `allow_failures` on PHP 7.2
- removed `composer self-update`

/cc @ezimuel

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,8 +69,6 @@ matrix:
     - php: 7.2
       env:
         - DEPS=latest
-  allow_failures:
-    - php: 7.2
 
 before_install:
   - if [[ $TEST_COVERAGE != 'true' ]]; then phpenv config-rm xdebug.ini || return 0 ; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,6 @@ matrix:
 
 before_install:
   - if [[ $TEST_COVERAGE != 'true' ]]; then phpenv config-rm xdebug.ini || return 0 ; fi
-  - travis_retry composer self-update
 
 install:
   - travis_retry composer install $COMPOSER_ARGS --ignore-platform-reqs
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,22 +15,12 @@ env:
   global:
     - COMPOSER_ARGS="--no-interaction"
     - COVERAGE_DEPS="satooshi/php-coveralls"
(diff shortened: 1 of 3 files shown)


ezimuel:
@webimpress since PHP 7.2 has not been yet released, we can leave the `allow_failures` until that and fix the Travis [issue](https://travis-ci.org/zendframework/zend-db/builds/306347111?utm_source=github_status&utm_medium=notification).

webimpress:
@ezimuel please see #276 for PHP 7.2 compatibility fixes.

I think we can remove `allow_failures` on PHP 7.2 because the release is just around the corner, we already have RC6 and stable release should be next Thursday. If something is still gonna break on this very late RC probably then we will have the same issue on the final version, unless you know about some bugs in PHP 7.2 which causes failing of zend-db?

I'd like to add that we've already removed `allow_failures` in many zf repositories.

ezimuel:
@webimpress it seems there's an issue with phpunit [here](https://travis-ci.org/zendframework/zend-db/jobs/306347128#L648) with PHP 7.2. All the other PHP versions are ok..

webimpress:
@ezimuel what is the point of supporting still PHP 5.5? It causes problem with update PHPUnit because version 5.7 supports only PHP 5.6+ and version 6.4 - PHP 7.0+. If we'd like to run tests here across all PHP versions 5.5-7.2 we need add support for PHPUnit 4.8 || 5.7 (at least), which is quite difficult and pointless as we are going to drop it soon (hopefully).
Maybe better then drop PHP 5.5, 5.6, 7.0 as was announced in https://framework.zend.com/blog/2017-06-06-zf-php-7-1.html, update PHPUnit to 6.4 and start adding features with PHP 7.1 compatibility, to move forward with this library.... I can't remember any other zf library which still support PHP 5.5...
It's more important to support all new PHP versions than the old one, IMHO.

ezimuel:
@webimpress I spoke with @weierophinney and we decided to support PHP 5.6. We can remove the support of PHP 5.5 and add PHP 7.2, since it will be released in a few days.

webimpress:
@ezimuel do you want me remove PHP 5.5 support in this PR?

ezimuel:
@webimpress yes, please. Thanks!

webimpress:
@ezimuel it's updated now, let's wait for tests results before merging

webimpress:
@ezimuel it can be merged now, tests fails on PHP 7.2 with lowest deps, as the issue with PHPUnit. I'll work on updating PHPUnit in separate PR soon. Thanks!

ezimuel:
@webimpress thanks!

PR #283, opened 2017-11-23
Fixed composer scripts

webimpress:
The repo is still using old CS library and tools

/cc @ezimuel


ezimuel:
@webimpress Thanks!

PR #190, opened 2016-11-13
New coding standard

Koopzington:
Usage of the newly introduced zendframework/zend-coding-standard for CS checks.

PR #284, opened 2017-11-28
New coding standard - follow up to #190

webimpress:
Based on PR #190, `master` merged into and all conflicts resolved


webimpress:
@ezimuel it can be merged now, and #190 should be closed.

Then I would be able work on PHPUnit update.

PR #287, opened 2017-11-30
Hotfix #285

### LINKED ISSUE #285: Potential issue on master - after merging #276
As noted to @ezimuel on Slack we have potential issue here:
https://github.com/zendframework/zend-db/commit/bef0d5ab124a6176e239d3fdac588a75cc2585a0#diff-1b78954777bbbf9668cb36aef77507c2R248 in case of `else` (after merging PR #276).

It should be verified by unit tests and fixed.

/cc @remicollet

webimpress:
I've checked and changing to `else` is good enough solution. Even if we have there non-string type it's gonna work as before, see: 
https://wiki.php.net/rfc/counting_non_countables\

> Calling `count()` on a scalar or object that doesn't implement the Countable interface returns 1.

Removed also part of the condition to check if variable is instance of Countable. It is not possible in that place. Please note the condition was wrong and not even working because `Countable` was not imported.

Resolves #285 

/cc @ezimuel @remicollet

### DIFF
--- a/src/Sql/AbstractSql.php
+++ b/src/Sql/AbstractSql.php
@@ -243,7 +243,7 @@ abstract class AbstractSql implements SqlInterface
                 foreach ($paramsForPosition as $multiParamsForPosition) {
                     if (is_array($multiParamsForPosition) || $multiParamsForPosition instanceof Countable) {
                         $ppCount = count($multiParamsForPosition);
-                    } elseif (is_string($multiParamsForPosition)) {
+                    } else {
                         $ppCount = 1;
                     }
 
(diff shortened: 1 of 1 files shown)


shaunfreeman:
I was getting this warning message in PHP 7.2
`Warning: count(): Parameter must be an array or an object that implements Countable in /vendor/zendframework/zend-db/src/Sql/AbstractSql.php on line 244`
When using a reultset with Zend\DB\Resultset\Resultset this hotfix fixes this issue.

webimpress:
@shaunfreeman the fix will be released with 2.9.0, probably soon :)

ezimuel:
@webimpress thanks! @shaunfreeman we'll release 2.9.0 next week.

Issue #288, opened 2017-12-07
DbAdapter query parameters in PDO


ArtemChornovol:
$dbAdapter->query('SET @@session.time_zone = :tz',
            [
                ':tz' => $timeZoneName,
            ]);

After https://github.com/zendframework/zend-db/pull/224, this line throws an error "Invalid parameter number: parameter was not defined"

ArtemChornovol:
This method changes param name before binding, but do not change it in query string


/**
     * @param string $name
     * @param string|null $type
     * @return string
     */
    public function formatParameterName($name, $type = null)
    {
        if ($type === null && ! is_numeric($name) || $type == self::PARAMETERIZATION_NAMED) {
            // using MD5 because of the PDO restriction [A-Za-z0-9_] for bindParam name
            return ':' . md5($name);
        }

        return '?';
    }

Ocramius:
@ArtemChornovol scenario above needs tests - not sure what `formatParameterName` is supposed to help with here.

ezimuel:
@ArtemChornovol damn, you right! The #224 introduced an issue, I completely failed to fix #35 not considering the params in the original SQL statement. This story is the proof that we need more tests for zend-db!!! I need to revert #224 and provide a new release 2.9.1. I'm working on it.

PR #289, opened 2017-12-07
Revert PR #224 due to issue like #288

ezimuel:
This PR revert #224 due to issue like #288. 

**Summary:** I tried to solve #35 with #224 but I didn't cover all the use cases, for instance #288. After reading also the php-cvs [46848](http://news.php.net/php.cvs/46848) and [47302](http://news.php.net/php.cvs/47302), mentioned in PHP BUG [43130](https://bugs.php.net/bug.php?id=43130), I decided to revert #224 and mark #35 as won't fix. It's not considered a best practice to use these characters in bind param name.

### DIFF
--- a/src/Adapter/Driver/Pdo/Pdo.php
+++ b/src/Adapter/Driver/Pdo/Pdo.php
@@ -304,8 +304,7 @@ class Pdo implements DriverInterface, DriverFeatureInterface, Profiler\ProfilerA
     public function formatParameterName($name, $type = null)
     {
         if ($type === null && ! is_numeric($name) || $type == self::PARAMETERIZATION_NAMED) {
-            // using MD5 because of the PDO restriction [A-Za-z0-9_] for bindParam name
-            return ':' . md5($name);
+            return ':' . $name;
         }
 
         return '?';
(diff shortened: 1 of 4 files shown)


weierophinney:
I guess the question I have is: do we want to safeguard against invalid parameter names when using PDO? #224 provided a fix for that, clearly, but as you also note, a number of other bugs indicate that most recommend against using invalid characters in the first place.

Perhaps we could do a check for invalid characters (basically, anything outside the accepted character set) and raise an exception?

webimpress:
@weierophinney 
> Perhaps we could do a check for invalid characters (basically, anything outside the accepted character set) and raise an exception?

That's actually a good idea.

@ezimuel 
I'm not sure if this PR completely reverting #224, I can see here "Files changed 4 (+8 −37)", but in #224 we have "Files changed 6 (+67 −5)"

ezimuel:
@weierophinney throwing exception is a good idea, I'll work on it.

@webimpress yes, it reverts #224 completely, I checked again. #224 contains multiple changes on the same code base.

ezimuel:
@weierophinney, @webimpress I added the Exception for invalid param name.

ezimuel:
@weierophinney thanks!

Issue #262, opened 2017-08-01
Few `limit` issue


autowp:
Briefly: SQL queries fails when few `limit` expressions is used (sub-select, maybe unions)

Context: PDO Mysql driver 

Expected SQL (simplified):
```sql
SELECT (SELECT column FROM foo LIMIT 1) AS column 
FROM bar 
LIMIT 99
```

PHP code to build them
```php
$subSelect = new Select('foo');
$subSelect->columns(['column'])->limit(1);

$select = new Select('bar');
$select->columns(['column' => $subSelect])->limit(99);

$statement = $sql->prepareStatementForSqlObject($select);
print_r($statement);
```

Output (simplified)
```
[sql:protected] => SELECT (SELECT `foo`.`column` AS `column` FROM `foo` LIMIT :limit) AS `column` FROM `bar` LIMIT :limit
[parameterContainer:protected] => Zend\Db\Adapter\ParameterContainer Object
        (
            [data:protected] => Array
                (
                    [limit] => 99
                )

            [positions:protected] => Array
                (
                    [0] => limit
                )
        )
```

Same value sent to mysql for both `limit` statements:

```sql
SELECT (SELECT column FROM foo LIMIT 99) AS column 
FROM bar 
LIMIT 99
```

Test: https://github.com/autowp/zend-db/commit/95efe84a618b3a08bae134897a434852e2587651

The only workaround is to create sub-select with SQL-string

michalbundyra:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/68.

PR #299, opened 2018-01-18
Update SelectDecorator.php

tptrixtop:
Add support for multi limit in nested queries;
That fix need for nested queries inside field parameters;

Fix bug for nested queries inside field parameters;

Provide a narrative description of what you are trying to accomplish:

PR #300, opened 2018-01-19
Add support for nested query in field params

tptrixtop:
Sorry, previous request was incorrect.

This one is correct.
Add support for multi limit in nested queries;
That fix need for nested queries inside field parameters;

In previous thread was mentioned that this is fix for that issue: #262 

Fix bug for nested queries inside field parameters;

### DIFF
--- a/src/Sql/Platform/Mysql/SelectDecorator.php
+++ b/src/Sql/Platform/Mysql/SelectDecorator.php
@@ -50,7 +50,8 @@ class SelectDecorator extends Select implements PlatformDecoratorInterface
             return;
         }
         if ($parameterContainer) {
-            $parameterContainer->offsetSet($this->processInfo['paramPrefix'] . 'limit', $this->limit, ParameterContainer::TYPE_INTEGER);
+            $param_prefix = $this->processInfo['paramPrefix'];
+            $parameterContainer->offsetSet($param_prefix . 'limit', $this->limit, ParameterContainer::TYPE_INTEGER);
             return [$driver->formatParameterName('limit')];
         }
 
@@ -66,7 +67,8 @@ class SelectDecorator extends Select implements PlatformDecoratorInterface
             return;
         }
         if ($parameterContainer) {
-            $parameterContainer->offsetSet('offset', $this->offset, ParameterContainer::TYPE_INTEGER);
+            $param_prefix = $this->processInfo['paramPrefix'];
+            $parameterContainer->offsetSet($param_prefix . 'offset', $this->offset, ParameterContainer::TYPE_INTEGER);
             return [$driver->formatParameterName('offset')];
         }
 
--- a/src/Sql/Platform/Mysql/SelectDecorator.php
+++ b/src/Sql/Platform/Mysql/SelectDecorator.php
@@ -50,8 +50,8 @@ class SelectDecorator extends Select implements PlatformDecoratorInterface
             return;
         }
         if ($parameterContainer) {
-            $param_prefix = $this->processInfo['paramPrefix'];
-            $parameterContainer->offsetSet($param_prefix . 'limit', $this->limit, ParameterContainer::TYPE_INTEGER);
+            $paramPrefix = $this->processInfo['paramPrefix'];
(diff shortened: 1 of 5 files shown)


webimpress:
@tptrixtop as @froschdesign mentioned before in https://github.com/zendframework/zend-db/pull/299#issuecomment-358620450 can you add test case for it, please?

tptrixtop:
I have add tests for that case. 
And 1 more fix for the same issue with offset param in nested queries;

src/Sql/Platform/Mysql/SelectDecorator.php
```diff
@@ -50,7 +50,8 @@ protected function processLimit(
             return;
         }
         if ($parameterContainer) {
-            $parameterContainer->offsetSet('limit', $this->limit, ParameterContainer::TYPE_INTEGER);
+            $param_prefix = $this->processInfo['paramPrefix'];
```
webimpress: Please use camelCase variable names instead of underscore_separated style.

tptrixtop: Yes, i think case with 1 query is valid, i will add it)

tptrixtop:
Done. I have added test for single nested query ( few limit case #262 ).
Variable renamed from underscore to camel case.

test/Sql/Platform/Mysql/SelectDecoratorTest.php
```diff
@@ -95,10 +96,74 @@ public function dataProvider()
...
+            ->columns([
+                'cnt' => new Expression('count(foo1.id)')
+            ])->limit(0)->offset(5);
+
+        $select3 = new Select;
+        $select3->from('foo')
+            ->columns([
+                'res'  => $nestedSelect0,
+            ])
+            ->limit('10')->offset('5');
```
webimpress: I would set here different offset, now we have the same as in the nested query, and it's possible that something is getting mixed, so to make it clear I would change the value.

webimpress: Another question also here: why these values are passed as strings, not as integers?

test/Sql/Platform/Mysql/SelectDecoratorTest.php
```diff
@@ -95,10 +96,74 @@ public function dataProvider()
...
+                'cnt' => new Expression('count(foo2.id)')
+            ])->limit(50)->offset(101);
+
+        $select4 = new Select;
+        $select4->from('foo')
+            ->columns([
+                'res'  => $nestedSelect0,
+                'res0' => $nestedSelect1
+            ])
+            ->limit('10')->offset('5');
```
webimpress: Also here, I would change the offset, to be different than in nested queries.

tptrixtop:
I have just copied them from my working code)

tptrixtop:
I have remove extra quotes and change offset & limit.

I did not check it for other platforms. Only for MySQL. 
I will try to check it for other platforms)

tptrixtop:
Done.
Limit & Offset are working in Oracle, MS Sql & IbmDb2.

Only additional fix needed in Sqlite. I have fixed it & added test for that & committed;

src/Sql/Platform/Sqlite/SelectDecorator.php
```diff
@@ -55,6 +55,43 @@ protected function processStatementStart(
...
+            return;
+        }
+        if ($parameterContainer) {
+            $paramPrefix = $this->processInfo['paramPrefix'];
+            $parameterContainer->offsetSet($paramPrefix . 'offset', $this->offset, ParameterContainer::TYPE_INTEGER);
+            return [$driver->formatParameterName('offset')];
+        }
+
+        return [$this->offset];
+    }
```
webimpress: I'd just to note that these methods are exactly the same as in:
https://github.com/zendframework/zend-db/blob/4675b97a5991279313818145538b70558419e410/src/Sql/Platform/Mysql/SelectDecorator.php#L41-L76

and the difference with parent class `Select` is to not quote offset and limit values.
I think it's fine, for MySQL it was added in that commit: https://github.com/zendframework/zend-db/commit/91bfba7944e36702c79e9acba7ebd7022b9a4611

ezimuel:
@tptrixtop thanks!

PR #303, opened 2018-02-01
Fix #295: handle empty array as datasource at AbstractResultSet::initialize() at php 7.2

### LINKED ISSUE #295: Initializing a result set with empty array produces warning on PHP 7.2
This produces a warning on PHP 7.2:
```php
$input = array(); // May actually be the result of another operation
$resultSet = new \Zend\Db\ResultSet\ResultSet;
$resultSet->initialize($input);
```
> count(): Parameter must be an array or an object that implements Countable

in ResultSet/AbstractResultSet.php, line 80

Since initializing a ResultSet with an array is perfectly valid, an empty array should be accounted for.

samsonasik:

- [x] Are you fixing a bug?
            is `count()` on datasource whenever datasource is empty array 
            when datasource is empty array, it got error: "count(): Parameter must be an array or an object that implements Countable"
            check the `current()` method, when it return `false`, set the count itself as 0.
  - [x] Add a `CHANGELOG.md` entry for the fix.

Fix #295

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Fixed
 
-- Nothing.
+- [#303](https://github.com/zendframework/zend-db/pull/303) fix error when datasource passed to `AbstractResultSet::initialize()` is empty array at php 7.2 environment.
 
 ## 2.9.2 - 2017-12-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Fixed
 
-- [#303](https://github.com/zendframework/zend-db/pull/303) fix error when datasource passed to `AbstractResultSet::initialize()` is empty array at php 7.2 environment.
+- [#295](https://github.com/zendframework/zend-db/pull/295) fix error when datasource passed to `AbstractResultSet::initialize()` is empty array at php 7.2 environment.
 
 ## 2.9.2 - 2017-12-11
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -22,7 +22,7 @@ All notable changes to this project will be documented in this file, in reverse
 
 ### Fixed
 
-- Nothing.
+- [#303](https://github.com/zendframework/zend-db/pull/303) fix error when datasource passed to `AbstractResultSet::initialize()` is empty array at php 7.2 environment.
```
froschdesign: #303 is the PR but #295 is the bug report. You should use #295.

samsonasik: done ;), I've updated the reference link to use #295

sfortop: @froschdesign 
> #303 is the PR but #295 is the bug report. You should use #295.

Why?
All changelog use PR link not bug report.

samsonasik:
is there any chance to get it merged? Thank you.

adear11:
Just wondering if this will be merged into a release any time soon. I have an Apigility API that is using zend-db. This is blocking me from upgrading it to php 7.2.

samsonasik:
rebased.

rzjack:
It's good to use zend-db now with PHP 7.2!
Unfortunately this patch changed the return type behaviour of AbstractResultSet->current().
When the result set is empty (array has zero elements), the current() method returned null prior this bugfix. But now it returns an empty array (array() or []) in that case.

I noticed this as the persistence layer in my application stopped working in some cases. So iIf anyone else runs into similar problems, maybe this entry helps!

samsonasik:
@rzjack I created test at #313 and the `AbstractResultSet::current()` still return null ( correct ). I can't reproduce issue with the test at AbstractResultSet::current() with initialize with empty array. Please provide unit test to prove the changed behavior.

PR #304, opened 2018-02-05
Fix for #35 and new integration test

ezimuel:
This PR fixes issue #35 and proposes a new integration test suite for `zend-db` (using the [travis-ci services](https://docs.travis-ci.com/user/database-setup/), `mysql` at the moment).

The issue #35 has been solved removing the internal mapping of the field name with PDO parameter name, using a `c_$i` incremental naming convention.

The new test folder contains unit test (`test/unit`) and integration test (`test/integration`) in separate folders. The test databases for integration are created using a `ZendIntegrationTest\Db\IntegrationTetListener` file. This new approach simplify the travis-ci integration without the usage of the `*_fixtures.sh` files in `.ci` folder.


froschdesign:
ping @alextech

ezimuel:
ping also @dkmuir, @alshayed, @webimpress, @alextech involved in #35, #224, #289.

.travis.yml
```diff
@@ -10,9 +10,9 @@ env:
   global:
     - COMPOSER_ARGS="--no-interaction"
     - COVERAGE_DEPS="satooshi/php-coveralls"
-    - TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL_HOSTNAME=localhost
-    - TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL_PASSWORD=
-    - TESTS_ZEND_DB_ADAPTER_DRIVER_PGSQL_HOSTNAME=localhost
+
+services:
+  - mysql
```
alextech: Move services to individual configuration blocks.
This will allow:
1. Ability to configure specific multiple supported versions. Eg, PgSQL 10 is not BC with 9.6.
2. Prevent travis from running multiple conflicting engines/configurations per build. Harder to see what version failed, unknown consequences.

.travis.yml
```diff
@@ -53,6 +59,8 @@ matrix:
     - php: 7.2
       env:
         - DEPS=locked
+        - TEST_INTEGRATION=true
+        - TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL=true
```
alextech: Put service here. One block per DB.

test/integration/IntegrationTestListener.php
```diff
@@ -0,0 +1,87 @@
...
+            return;
+        }
+        printf("\nIntegration test ended.\n");
+
+        if (getenv('TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL')) {
+            $this->dropMysqlDatabase();
+        }
+    }
+
+    private function createMysqlDatabase($dbFile)
```
alextech: Keep to convention used around the codebase of having this in _Platform_ class. Will make possible adding other engines.

test/integration/IntegrationTestListener.php
```diff
@@ -0,0 +1,87 @@
...
+            'mysql:host=' . getenv('TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL_HOSTNAME'),
+            getenv('TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL_USERNAME'),
+            getenv('TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL_PASSWORD')
+        );
+        if (false === $this->pdo->exec(sprintf(
+            "CREATE DATABASE IF NOT EXISTS %s",
+            getenv('TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL_DATABASE')
+        ))) {
+            throw new Exception(sprintf(
+                "I cannot create the MySQL %s test database",
```
alextech: In exceptions, add display of actual error occurred with $this->pdo->errorInfo(). 

`print_r($this->pdo->errorInfo(), true)` gives minimum formatting.
I had this fail and couldn't tell what was wrong.

test/integration/IntegrationTestListener.php
```diff
@@ -0,0 +1,87 @@
...
+                getenv('TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL_DATABASE')
+            ));
+        }
+
+        $this->pdo->exec('USE ' . getenv('TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL_DATABASE'));
+
+        if (false === $this->pdo->exec(file_get_contents($dbFile))) {
+            throw new Exception(sprintf(
+                "I cannot create the table for %s database. Check the %s file. ",
```
alextech: As above in this exception.
But one for "Use" may not be needed, as i could not find ways it could fail for MySQL.

test/integration/TestAsset/mysql.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE IF NOT EXISTS test (
```
alextech: Rename file and path to "fixtures" to use more widely used terminology.

ezimuel:
@alextech I think we should have specific integration tests for all the db engine that we support. Even for special cases like #35, where we need to reproduce issues like #288 and https://github.com/zendframework/zend-db/pull/178#issuecomment-255564783.

> I also do not think separating engines for high level utility tests, such as TableGateway, is good. This again creates false positive passes when someone modifies TableGateway, writes test for it in MySQL and does not replicate the test to other platforms.

If you change `TableGateway` and write a **specific test** for MySQL you are testing the changes with a specific adapter, that maybe will fail even if the high level tests of `TableGateway` are passing. This is good to find specific issue or bad design. I don't see false positive here. Integration test and unit test are completely different and can be also executed as separated thing (that's why I split in two folder and created two commands: `composer test` and `composer test-integration`).

> ...If same abstraction test fails and needs engine-specific handling, then abstraction is badly designed at that point

Again, this is actually very good, because we can discover issue and bad design. Tests are useful to find mistakes, they don't prove that everything is ok (it's never ok as you know, even if you have 100% test coverage).

Anyway, I'm reviewing your comment and I'm glad that you improved the PR with https://github.com/ezimuel/zend-db/pull/1, thanks!

alextech:
>If you change TableGateway and write a specific test for MySQL you are testing the changes with a specific adapter, that maybe will fail even if the high level tests of TableGateway are passing.

Exactly.  So I did not mean that unit tests should be used instead of integration. I meant that integration should not be in an adapter specific namespace for Design Pattern implementations.

Design Pattern implementations, unlike direct adapter interaction have added responsibility of having same API work *simultaneously* on all engines. When you write TableGateway test in MySQL namespace that ends up not working for some reason in SqlServer, PostgreSQL as the documentation advertises, its a problem. As random example, https://github.com/zendframework/zend-db/pull/304/files#diff-56dd6fd011c00c8a8c735d14eed5c841R27  If that test passes for MySQL but not PostgreSQL, is that not a false positive? 

Is everyone going to be diligent to copy, https://github.com/zendframework/zend-db/pull/304/files#diff-56dd6fd011c00c8a8c735d14eed5c841R88 4 times to our 4 supported platforms to verify something as basic as $tableGateway->update() works as advertised?

That does not mean don't keep them as integration. I mean it should not be isolated to one namespace or manually copied to all namespaces, but live in top level integration namespace and reran automatically for each integration engine in travis yaml config.

ezimuel:
@alextech ok, now I see your point. I agree to move the `TableGateway` at top level. That said, we can still have some special cases where we need to test `TableGateway` only for a specific vendor, to reproduce am issue coming for a bad design or environment related.

alextech:
Yes, there are some integration tests for TableGateway that *are* engine specific, eg, using delimited path as a table name for Pgsql and SqlServer.  I just remembered that case, sorry! So yes to 
>we can still have some special cases where we need to test TableGateway only for a specific vendor, to reproduce am issue coming

Issue #302, opened 2018-01-22
Reconnect on lost DB connection with PDO_MYSQL


finalJustize:
I am using php for long running scripts as a RabbitMq message comsumer. Messages can have DB access implemented as ZF3 / zend-db Tablegateway. The DB connection used is PDO_MYSQL.
In case DB connection is killed (MYSQL_CONNECTION_TIMEOUT or other), AbstractTableGateway throws an error (HY000 - 2006 - MySQL server has gone away) when trying to execute a statement against the closed connection.

I expect that Tablegateway / Adapter / db Driver is smart enough to handle lost connection and reconnect or at least handle "unconnected" state and connect (again).

### Code to reproduce the issue

* use some Tablegateway connected via PDO_MYSQL in a long running loop
* while the loop runs, go to the MySQL-Server and kill the connection manually
* you will get "HY000 - 2006 - MySQL server has gone away" in Tablegateway / Adapter / db Driver. 

### Expected results

when the DB connection is lost the DB managing objects should be able to connect again to the DB again as they did initially.

### Actual results

* when the DB connection is lost the DB managing objects are not able to connect again. An error is thrown: "HY000 - 2006 - MySQL server has gone away" in Tablegateway / Adapter / db Driver.

michalbundyra:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/56.

PR #306, opened 2018-03-04
Reconnect on lost DB connection with PDO_MYSQL

tptrixtop:
I have added support for reconnect on connection lost for mysqli.

Fix issue #302


tptrixtop:
Sorry miss correct branch)

PR #308, opened 2018-03-13
use direct "hydrator" property call instead of "getHydrator()" at HydratingResultSet:: toArray()

samsonasik:
as call inside a class.


ezimuel:
@samsonasik thanks!

PR #311, opened 2018-04-03
Fixes integration test listener

ezimuel:
This PR fixes the integration test listener if the environment variables are not available.

### DIFF
--- a/test/integration/IntegrationTestListener.php
+++ b/test/integration/IntegrationTestListener.php
@@ -41,7 +41,7 @@ class IntegrationTestListener extends BaseTestListener
             $this->fixtureLoader = new \ZendIntegrationTest\Db\Platform\PgsqlFixtureLoader();
         }
 
-        if (!isset($this->fixtureLoader)) {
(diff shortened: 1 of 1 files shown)


test/integration/IntegrationTestListener.php
```diff
@@ -42,12 +41,17 @@ public function startTestSuite(TestSuite $suite)
...
+        }
+        printf("\nIntegration test started.\n");
         $this->fixtureLoader->createDatabase();
     }
 
     public function endTestSuite(TestSuite $suite)
     {
-        if ($suite->getName() !== 'integration test') {
+        if ($suite->getName() !== 'integration test' ||
+            ! isset($this->fixtureLoader)) {
```
weierophinney: This should read:

```php
if ($suite->getName() !== 'integration test'
    || ! isset($this->fixtureLoader)
) {
```

(Operators go on the next line, so that we can add more conditions without requiring additional lines of changes.)

I'll fix this on merge.

PR #312, opened 2018-04-11
Change Return Type in selectWith() method

lowtower:
Change Return Type of selectWith() method

- [x] Are you fixing a bug?
> `phpstan` raises an error saying that `ResulSetInterface` has no method `current()`.
> The `selectWith()` method has defined the `@returnType` as `ResultSetInterface`.
> The method calls the method `executeSelect()`, which returns `ResultSet`.
> The return types should be the same.
> removes also the `@throws` flag as it doesn't throw an exception itself, but just the called methods.
  - [ ] Detail the new, expected behavior.
  - [ ] Base your feature on the `master` branch, and submit against that branch.
  - [ ] Add a regression test that demonstrates the bug, and proves the fix.

- [ ] Are you creating a new feature?
  - [ ] Why is the new feature needed? What purpose does it serve?
  - [ ] How will users use the new feature?
  - [ ] Base your feature on the `develop` branch, and submit against that branch.
  - [ ] Add only one feature per pull request; split multiple features over multiple pull requests
  - [ ] Add tests for the new feature.
  - [ ] Add documentation for the new feature.
  - [ ] Add a `CHANGELOG.md` entry for the new feature.

- [ ] Is this related to documentation?

### DIFF
--- a/src/TableGateway/AbstractTableGateway.php
+++ b/src/TableGateway/AbstractTableGateway.php
@@ -198,8 +198,7 @@ abstract class AbstractTableGateway implements TableGatewayInterface
 
     /**
      * @param Select $select
-     * @return ResultSetInterface
-     * @throws \RuntimeException
+     * @return ResultSet
      */
     public function selectWith(Select $select)
     {
--- a/src/TableGateway/AbstractTableGateway.php
+++ b/src/TableGateway/AbstractTableGateway.php
@@ -177,7 +177,7 @@ abstract class AbstractTableGateway implements TableGatewayInterface
      * Select
      *
      * @param Where|\Closure|string|array $where
-     * @return ResultSet
+     * @return AbstractResultSet
      */
     public function select($where = null)
     {
@@ -198,7 +198,7 @@ abstract class AbstractTableGateway implements TableGatewayInterface
 
     /**
      * @param Select $select
-     * @return ResultSet
+     * @return AbstractResultSet
      */
     public function selectWith(Select $select)
     {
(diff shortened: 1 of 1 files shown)


tptrixtop:
May be use then AbstractResultSet in lieu of ResultSet ?

samsonasik:
`selectWith()` can return `HydratingResultSet`, so `AbstractResultSet` is better.

lowtower:
Then `select()`, `selectWith()` and `executeSelect()` should all return `AbstractResultSet`.
Right?

froschdesign:
Why `AbstractResultSet`? Compare with:

https://github.com/zendframework/zend-db/blob/adfe190ed0e6558d83558f7f68abfd7ccfda1e26/src/TableGateway/AbstractTableGateway.php#L58-L61

https://github.com/zendframework/zend-db/blob/adfe190ed0e6558d83558f7f68abfd7ccfda1e26/src/TableGateway/AbstractTableGateway.php#L109-L111

https://github.com/zendframework/zend-db/blob/adfe190ed0e6558d83558f7f68abfd7ccfda1e26/src/TableGateway/AbstractTableGateway.php#L241-L248

samsonasik:
missing use statement:
```
use Zend\Db\ResultSet\AbstractResultSet;
```

samsonasik:
@froschdesign how about using `@var ResultSetInterface| AbstractResultSet` for `$resultSetPrototype` property ?

lowtower:
@froschdesign 
indeed, not consistant!
What do you propose?

ezimuel:
@samsonasik `Zend\Db\ResultSet\AbstractResultSet` implements `Zend\Db\ResultSet\ResultSetInterface`, it's better to use this interface as return type.
@lowtower can you propose a new PR chaning all the types to `ResultSetInterface`? Thanks!
I'm closing this PR in favor of a new one.

PR #313, opened 2018-04-24
Test AbstractResultSet::current() to return null on empty array

samsonasik:
This is based on @rzjack comment at https://github.com/zendframework/zend-db/pull/303#issuecomment-383976062

  - [x] Are you fixing a bug?
           initialize `AbstractResultSet` with empty array.
           The `AbstractResultset::current()` return `[]`.
           The `AbstractResultset::current()` return null. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


samsonasik:
@rzjack I can't reproduce issue with the test at `AbstractResultSet::current()` with initialize with empty array.

PR #315, opened 2018-05-08
count() usage optimization

samsonasik:


src/TableGateway/Feature/MetadataFeature.php
```diff
@@ -73,11 +73,11 @@ public function postInitialize()
             throw new Exception\RuntimeException('A primary key for this column could not be found in the metadata.');
         }
 
-        if (count($pkc->getColumns()) == 1) {
-            $pkck = $pkc->getColumns();
+        $pkck = $pkck->getColumns();
+        if (count($pkc) == 1) {
```
webimpress: We can use `===` here now.

samsonasik: done ;)

PR #324, opened 2018-07-03
remove unneeded strtolower() call as previously already called

samsonasik:

In previous line, there is already `$value = strtolower((string) $value);`


ezimuel:
@samsonasik Thanks!

Issue #318, opened 2018-05-16
MySQL SelectDecorator - subselect limit parameter wrong name


smartwebproducts:
- [x] I was not able to find an [open](https://github.com/zendframework/zend-db/issues?q=is%3Aopen) or [closed](https://github.com/zendframework/zend-db/issues?q=is%3Aclosed) issue matching what I'm seeing.
 - [x] This is not a question. (Questions should be asked on [slack](https://zendframework.slack.com/) ([Signup for Slack here](https://zendframework-slack.herokuapp.com/)) or our [forums](https://discourse.zendframework.com/).)

When you have a FROM clause that includes a query with a LIMIT in it, and you try to prepare the query, you get following error:

```
Fatal error: Uncaught PDOException: SQLSTATE[HY093]: Invalid parameter number: parameter was not defined in /Users/jkeppens/Documents/Projects/Connecting-Expertise/dev/api.connecting-expertise.com/vendor/zendframework/zend-db/src/Adapter/Driver/Pdo/Statement.php:239
Stack trace:
#0 /Users/jkeppens/Documents/Projects/Connecting-Expertise/dev/api.connecting-expertise.com/vendor/zendframework/zend-db/src/Adapter/Driver/Pdo/Statement.php(239): PDOStatement->execute()
```

The reason for this is because the parameter name for the subselect's limit is `:limit` while in the parameterContainer it is `:subselect1limit`.

This problem is rectified by modifying `line 55` of  `src/Sql/Platform/Mysql/SelectDecorator.php` from:

```php
return [$driver->formatParameterName('limit')];
```
to

```php
return [$driver->formatParameterName($paramPrefix . 'limit')];
```

tptrixtop:
Looks like that was fixed by this PR #300 )

michalbundyra:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/50.

PR #329, opened 2018-07-23
Fixes Issue #318 - SelectDecorator - subselect limit parameter not prefixed

fuel-odrais:

- [X] Are you fixing a bug?
  - [X] Detail how the bug is invoked currently.
         Call `prepareStatementForSqlObject` on a `Select` with a sub-`Select` that has `limit` and/or `offset` set
  - [X] Detail the original, incorrect behavior.
         A PDO Exception is thrown: `PDOException: SQLSTATE[HY093]: Invalid parameter number: parameter was not defined`
  - [X] Detail the new, expected behavior.
        The statement is properly generated.
  - [X] Base your feature on the `master` branch, and submit against that branch.
  - [X] Add a regression test that demonstrates the bug, and proves the fix.
  - [X] Add a `CHANGELOG.md` entry for the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,9 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Fixed
 
-- Nothing.
+- [#329](https://github.com/zendframework/zend-db/pull/329) fix Exception
+ thrown when calling prepareStatementForSqlObject on a Select with a
+ sub-Select that has limit and/or offset set
 
 ## 2.9.3 - 2018-04-09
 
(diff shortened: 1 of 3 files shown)


tptrixtop:
Looks like that was fixed by that #300

fuel-odrais:
@tptrixtop:

In PR #300, src/Sql/Platform/Mysql/SelectDecorator.php line 54 is this:

`$parameterContainer->offsetSet($paramPrefix . 'limit', $this->limit, ParameterContainer::TYPE_INTEGER);`

That code associates the Select's `limit` property with a SQL parameter named `$paramPrefix` . 'limit'.

Then we return the name of the SQL parameter that will actually be returned from the generated query. That string value is missing the `$paramPrefix` value.

I've included unit tests to demonstrate the bug and would be happy to make a commit with the tests only so you can verify.

tptrixtop:
@fuel-odrais 

Ah, got it sir)

tptrixtop:
ping @webimpress i think)

webimpress:
@tptrixtop Sorry, I am not a maintainer... Ping @ezimuel :)

ezimuel:
@fuel-odrais Thanks!

MatyCZ:
When will be this fix released as new zend-db version?

PR #333, opened 2018-08-08
Added IbmDb2 db2_pclose to disconnect

JargonMan:
Updated the disconnect method to choose db2_pclose or db2_close based on the resource's connectionParameter for persistence.

The public function connect() creates either persistent or non persistent db2 connections based on a connection parameter.

This update uses the same logic in the public function disconnect() to close the connection.  Previously persistent connections would not have been closed with the db2_close() call, because they need the db2_pclose() call.

### DIFF
--- a/src/Adapter/Driver/IbmDb2/Connection.php
+++ b/src/Adapter/Driver/IbmDb2/Connection.php
@@ -149,7 +149,28 @@ class Connection extends AbstractConnection
     public function disconnect()
     {
         if ($this->resource) {
-            db2_close($this->resource);
+            
+            // localize
+            $p = $this->connectionParameters;
+
+            // given a list of key names, test for existence in $p
+            $findParameterValue = function (array $names) use ($p) {
+                foreach ($names as $name) {
+                    if (isset($p[$name])) {
+                        return $p[$name];
+                    }
+                }
+    
+                return;
+            };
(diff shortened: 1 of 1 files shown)


src/Adapter/Driver/IbmDb2/Connection.php
```diff
@@ -149,7 +149,28 @@ public function isConnected()
     public function disconnect()
     {
         if ($this->resource) {
-            db2_close($this->resource);
+            
+            // localize
+            $p = $this->connectionParameters;
+
+            // given a list of key names, test for existence in $p
+            $findParameterValue = function (array $names) use ($p) {
```
Ocramius: Make a new private method returning `bool` instead. Also, `array_change_key_case()` is probably a better approach, since nothing guarantees that mIxEDCasE won't be given as parameter.

Ocramius:
Was this supposed to be closed?

JargonMan:
Sorry, I'm new at GitHub and thought the pull request was rejected and I needed to create a new one.

Ocramius:
Hah, ok. You can just send more commits to this branch, and that will
update the patch.

weierophinney:
This repository has been closed and moved to laminas/laminas-db; a new issue has been opened at https://github.com/laminas/laminas-db/issues/27.

weierophinney:

PR #334, opened 2018-08-09
WIP: Fix 330, return NULL if current() is not countable

### LINKED ISSUE #330: Zend\Db\ResultSet\ResultSet->current() returns false instead of NULL
This problem was there in version 2.7. It was fixed in 2.8, but came back in 2.9.

- Database: 10.0.35-MariaDB
- PHP version: 7.2.7
- Zend Db version: 2.9.3

To reproduce, use a database with some empty table, and execute:

```
$myAdapter = new \Zend\Db\Adapter\Adapter([
    'host'      => 'mariadb',
    'driver'    => 'Mysqli',
    'username'  => 'devppl',
    'password'  => '******',
    'database'  => 'mydatabase',
    'charset'   => 'utf8',
    'options'   => ['buffer_results' => true]
]);

$tableGateway = new \Zend\Db\TableGateway\TableGateway('myEmptyTable', $myAdapter);
$row = $tableGateway->select('myField = 999')->current();
var_export($row);
```

Expected result would be NULL, I get false instead.

I noticed that when I set buffer_results to false, I don't have the bug.

ezimuel:
This PR fixes #330 returning NULL if current() result is not countable. It removes and changes part of the `Zend\Db\ResultSet\AbstractResultSet::current()` implementation modified with https://github.com/zendframework/zend-db/commit/8276cf6cfe45dce834979f0e714eea81065a7493.

### DIFF
--- a/src/ResultSet/AbstractResultSet.php
+++ b/src/ResultSet/AbstractResultSet.php
@@ -194,11 +194,6 @@ abstract class AbstractResultSet implements Iterator, ResultSetInterface
      */
     public function current()
     {
-        if (-1 === $this->buffer) {
-            // datasource was an array when the resultset was initialized
-            return $this->dataSource->current();
-        }
-
         if ($this->buffer === null) {
             $this->buffer = -2; // implicitly disable buffering from here on
         } elseif (is_array($this->buffer) && isset($this->buffer[$this->position])) {
@@ -208,7 +203,7 @@ abstract class AbstractResultSet implements Iterator, ResultSetInterface
         if (is_array($this->buffer)) {
             $this->buffer[$this->position] = $data;
         }
-        return is_array($data) ? $data : null;
+        return (is_array($data) || $data instanceof Countable) ? $data : null;
     }
 
     /**


ezimuel:
Adding @bartmcleod to review this, since I changed the previous implementation https://github.com/zendframework/zend-db/commit/8276cf6cfe45dce834979f0e714eea81065a7493

src/ResultSet/AbstractResultSet.php
```diff
@@ -208,7 +203,7 @@ public function current()
         if (is_array($this->buffer)) {
             $this->buffer[$this->position] = $data;
         }
-        return is_array($data) ? $data : null;
+        return (is_array($data) || $data instanceof Countable) ? $data : null;
```
Ocramius: Why is `Countable` a valid type, but (for example) `Traversable` not? What's the rationale behind the change?

src/ResultSet/AbstractResultSet.php
```diff
@@ -194,11 +194,6 @@ public function key()
      */
     public function current()
     {
-        if (-1 === $this->buffer) {
-            // datasource was an array when the resultset was initialized
-            return $this->dataSource->current();
-        }
-
```
bartmcleod: I had my doubts about this part, but at the time, it was needed, so I am surprised the Travis build succeeds without these lines. If it does, it is probably ok.

src/ResultSet/AbstractResultSet.php
bartmcleod: Since this apparently fixes #330 because of a bug I might have introduced earlier, it is not needed to cover every single interface that could be returned by ->current(). If it fixes #330 and does not break anything, then it is ok, but we should see a unit test that confirms #330 and thus explains the need for the fix.

ezimuel: @bartmcleod The code that you introduced [here](https://github.com/zendframework/zend-db/blob/8276cf6cfe45dce834979f0e714eea81065a7493/src/ResultSet/AbstractResultSet.php#L197-L200) just returns the current value of `dataSource` that can be a different type, depending on the adapter (e.g. for Mysqli is [false](https://github.com/zendframework/zend-db/blob/master/src/Adapter/Driver/Mysqli/Result.php#L56), for PDO is [null](https://github.com/zendframework/zend-db/blob/master/src/Adapter/Driver/Pdo/Result.php#L53), etc). Maybe, I should align all the default value of the adapter to be the same but I'm not sure this will introduce BC breaks in some existing project. That's why I focused my fix on the `AbstractResultSet`.

@Ocramius I'm trying to accomodate different return type for `Resultset::current()` coming from different adapter, it's seems there is inconsistency with different adapters (see above). Anyway, I'm still testing all the use cases, it's quite difficult without type hinting...

bartmcleod: @ezimuel would it be okay to check with `empty()`  and to always return null in that case? So that false and null are treated equally.

ezimuel:
Closing in favor of #337

PR #335, opened 2018-08-09
Disconnect persistent IbmDb2 connection

JargonMan:
Persistent database connections created with db2_pconnect don't close with the current disconnect method.  This change adds a callable method for closing persistent database connections by using db2_pclose().

### DIFF
--- a/src/Adapter/Driver/IbmDb2/Connection.php
+++ b/src/Adapter/Driver/IbmDb2/Connection.php
@@ -156,6 +156,21 @@ class Connection extends AbstractConnection
         return $this;
     }
 
+     /**
+     * Disconnect a persistent resource
+     *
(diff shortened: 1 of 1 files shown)


src/Adapter/Driver/IbmDb2/Connection.php
```diff
@@ -156,6 +156,21 @@ public function disconnect()
         return $this;
     }
 
+     /**
+     * Disconnect a persistent resource
+     *
+     * @return ConnectionInterface
+     */
+    public function disconnectPersistent()
```
Ocramius: Providing a completely different API for something that is an implementation detail is incorrect, and breaks LSP (expanding the interface for an adapter means that it is no longer an adapter)

PR #336, opened 2018-08-11
Add insert ignore

samsonasik:
- [x] Are you creating a new feature?
  - [x] Why is the new feature needed? What purpose does it serve?
            Useful to ignore duplicate entry error.
  - [x] How will users use the new feature?

```php
$insert= new InsertIgnore('tablename');
$insert->values($data);

$statement = $sql->prepareStatementForSqlObject($insert);
$statement->execute();
```
  - [x] Base your feature on the `develop` branch, and submit against that branch.
  - [x] Add only one feature per pull request; split multiple features over multiple pull requests
  - [x] Add tests for the new feature.
  - [x] Add a `CHANGELOG.md` entry for the new feature.

### DIFF
--- a/src/Sql/InsertIgnore.php
+++ b/src/Sql/InsertIgnore.php
@@ -1,0 +1,21 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Db\Sql;
+
+class InsertIgnore extends Insert
+{
+    /**
+     * @var array Specification array
+     */
+    protected $specifications = [
+        self::SPECIFICATION_INSERT => 'INSERT IGNORE INTO %1$s (%2$s) VALUES (%3$s)',
+        self::SPECIFICATION_SELECT => 'INSERT IGNORE INTO %1$s %2$s %3$s',
+    ];
+}
--- a/test/unit/Sql/InsertIgnoreTest.php
+++ b/test/unit/Sql/InsertIgnoreTest.php
@@ -34,7 +34,7 @@ class InsertIgnoreTest extends TestCase
     }
 
     /**
-     * @covers \Zend\Db\Sql\Insert::into
+     * @covers \Zend\Db\Sql\InsertIgnore::into
      */
     public function testInto()
     {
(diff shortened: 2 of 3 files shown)


src/Sql/InsertIgnore.php
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Zend\Db\Sql;
+
+class InsertIgnore extends Insert
+{
+    /**
+     * @var array Specification array
+     */
+    protected $specifications = [
+        self::SPECIFICATION_INSERT => 'INSERT IGNORE INTO %1$s (%2$s) VALUES (%3$s)',
```
Ocramius: Are all supported drivers capable of evaluating `INSERT IGNORE`?

src/Sql/InsertIgnore.php
```diff
@@ -0,0 +1,21 @@
...
+ * @copyright Copyright (c) 2005-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Db\Sql;
+
+class InsertIgnore extends Insert
+{
+    /**
+     * @var array Specification array
```
Ocramius: Please specify the type of the array

test/unit/Sql/InsertIgnoreTest.php
```diff
@@ -0,0 +1,412 @@
...
+     * Sets up the fixture, for example, opens a network connection.
+     * This method is called before a test is executed.
+     */
+    protected function setUp()
+    {
+        $this->insert = new InsertIgnore;
+    }
+
+    /**
+     * @covers \Zend\Db\Sql\InsertIgnore::into
```
Ocramius: I suggest dropping all these `@covers`, ans only adding one at class level

src/Sql/InsertIgnore.php
samsonasik: probably MySQL only, that's why I didn't add `insertIgnore()` function to `Sql` and `AbstractTableGateway`

src/Sql/InsertIgnore.php
samsonasik: it is copied originally from `Insert` class, do you have sample of "type of the array"?

test/unit/Sql/InsertIgnoreTest.php
samsonasik: removed `@covers`

src/Sql/InsertIgnore.php
Ocramius: `string[]` in this case, or even `array<string, string>` (unsure if the tooling craps out with that, though)

src/Sql/InsertIgnore.php
Ocramius: Hmm, can we please survey that before introducing it then?

michalbundyra:
Thanks, @samsonasik!

PR #337, opened 2018-08-17
Fixes #330 current NULL for mysqli

### LINKED ISSUE #330: Zend\Db\ResultSet\ResultSet->current() returns false instead of NULL

ezimuel:
This PR fixes #330 and replaces #334. I focused the fix directly on mysqli adapter, and not anymore in the `Zend\Db\ResultSet\AbstractResultSet::current()`, changing the default `$currentData` value of `Zend\Db\Adapter\Driver\Mysqli\ResultSet` to NULL. I added an integration test to reproduce the issue, as reported in #330.

### DIFF
--- a/src/Adapter/Driver/Mysqli/Result.php
+++ b/src/Adapter/Driver/Mysqli/Result.php
@@ -51,9 +51,9 @@ class Result implements
     protected $nextComplete = false;
 
     /**
-     * @var bool
+     * @var mixed
      */
-    protected $currentData = false;
+    protected $currentData = null;
 
     /**
      *
--- a/test/integration/Adapter/Driver/Mysqli/TraitSetup.php
+++ b/test/integration/Adapter/Driver/Mysqli/TraitSetup.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-db for the canonical source repository
+ * @copyright Copyright (c) 2005-2018 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://github.com/zendframework/zend-db/blob/master/LICENSE.md New BSD License
+ */
+
+namespace ZendIntegrationTest\Db\Adapter\Driver\Mysqli;
+
+trait TraitSetup
+{
+    protected $variables = [
+        'hostname' => 'TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL_HOSTNAME',
+        'username' => 'TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL_USERNAME',
+        'password' => 'TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL_PASSWORD',
+        'database' => 'TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL_DATABASE',
+    ];
+
+    /**
+     * Sets up the fixture, for example, opens a network connection.
+     * This method is called before a test is executed.
+     */
+    protected function setUp()
+    {
+        if (! getenv('TESTS_ZEND_DB_ADAPTER_DRIVER_MYSQL')) {
+            $this->markTestSkipped('Mysqli integration test disabled');
+        }
+
+        if (! extension_loaded('mysqli')) {
+            $this->fail('The phpunit group integration-mysqli was enabled, but the extension is not loaded.');
+        }
+
+        foreach ($this->variables as $name => $value) {
+            if (! getenv($value)) {
+                $this->markTestSkipped(sprintf(
+                    'Missing required variable %s from phpunit.xml for this integration test',
+                    $value
+                ));
+            }
+            $this->variables[$name] = getenv($value);
+        }
+    }
+}
--- a/test/integration/Adapter/Driver/Mysqli/ConnectionTest.php
+++ b/test/integration/Adapter/Driver/Mysqli/ConnectionTest.php
@@ -12,37 +12,7 @@ use Zend\Db\Adapter\Driver\Mysqli\Connection;
 class ConnectionTest extends TestCase
 {
 
-    protected $variables = [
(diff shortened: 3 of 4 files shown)


ezimuel:
@bartmcleod I revert the change of #334, your commit https://github.com/zendframework/zend-db/commit/8276cf6cfe45dce834979f0e714eea81065a7493#diff-86b3a79320eb564e46456a1b75003271R197 has been reinsert, now the solution is directly in the mysqli adapter.

willy0275:
Good work. I'm just wondering if ResultSet/AbstractResultSet-->current() should not at least *enforce* that the resulting value is null or an array, or else it would throw an exception? We use null in the method signature but we actually don't enforce it? Seems a bit fishy to me.

willy0275:
(The idea of a ResultSet and pretty much the whole Zend\Db is to abstract the driver you're using, so "programming errors" in some drivers should not lead to different result to the current() method on ResultSet)

willy0275:
The reason I'm picky on this is that there's literally hundreds of places in my code where I do if ($row === null) to know if it's empty after calling ResultSet-->current(), so I feel the bug could come back any time in the future *and* I would not have a clear explanation of the error, the code keeps on going. At least, an exception "The driver is returning an invalid value" or something like that would allow me to know right away it's an internal error to Zend\Db.

ezimuel:
@willy0275 you right, there are some inconsistencies in these adapters return types. We can think to uniform everything and introduces Exceptions in ver. 3.0.0 since this will be a BC break.

willy-opale:
Any schedule on this important bug fix? We're still using a private repo which is a copy of zend db 2.9 with a temporary fix in the source code.

src/Adapter/Driver/Mysqli/Result.php
```diff
@@ -51,9 +51,9 @@ class Result implements
     protected $nextComplete = false;
 
     /**
-     * @var bool
+     * @var mixed
```
Ocramius: Do we have any stricter type that can be used here?

src/Adapter/Driver/Mysqli/Result.php
```diff
@@ -51,9 +51,9 @@ class Result implements
...
      */
-    protected $currentData = false;
+    protected $currentData = null;
```
Ocramius: This is a BC break for anything inheriting from this class. Sadly, the class is not `final`, so these changes land in future releases (`develop`)

test/integration/Adapter/Driver/Mysqli/TraitSetup.php
```diff
@@ -0,0 +1,43 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-db for the canonical source repository
+ * @copyright Copyright (c) 2005-2018 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://github.com/zendframework/zend-db/blob/master/LICENSE.md New BSD License
+ */
+
+namespace ZendIntegrationTest\Db\Adapter\Driver\Mysqli;
+
+trait TraitSetup
```
Ocramius: Let's please not use traits for stashing out duplicated code: this is static code, and should be kept as such (static utility)

src/Adapter/Driver/Mysqli/Result.php
ezimuel: This can be `null` or `array`, depending on the `current()` state. Moreover, I just reviewed all the adapter results and this is the scenario:
```
Adapter\Driver\Pgsql - current(): array|false
Adapter\Driver\Mysqli - current(): array|null
Adapter\Driver\Oci8 - current(): array|false
Adapter\Driver\Pdo - current(): array|object|false
Adapter\Driver\Sqlsrv - current(): array|false|null
```
I think we should try to normalize it. Maybe, we can throw an Exception if the adapter returns an error during the fetch process, and use `null` as empty value. Of course, this will be released with zend-db 3.0.0
@Ocramius and @willy0275 what do you think?

src/Adapter/Driver/Mysqli/Result.php
ezimuel: Yes, I'm planning it for zend-db 3.0.0. I'll point to `develop`, thanks!

test/integration/Adapter/Driver/Mysqli/TraitSetup.php
ezimuel: Can you elaborate why? Moreover, you mentioned this is static code?! Thanks.

src/Adapter/Driver/Mysqli/Result.php
ezimuel: I'm going to release this in 2.10.0 since it's a rollback of 2.8 behaviour.

ezimuel:
Merged in develop.

src/Adapter/Driver/Mysqli/Result.php
willy0275: Thanks! We are still on our local patched version of 2.9, so it's going to be great to be able to switch back to the main repo.

PR #338, opened 2018-09-07
restore missing use file

mistermann87:
This problem always appears when using AbstractTableGateway
In src/TableGateway/AbstractTableGateway.php:110 a new instance of ResultSet is created without a use statement.
Use statement added again - ResultSet can now be created correctly


ezimuel:
@mistermann87 Thanks! I added a unit test to cover the issue.

PR #340, opened 2018-10-15
Variable  was used before it was declared

henriquemattos:
Provide a narrative description of what you are trying to accomplish:

I think that this [commit](https://github.com/zendframework/zend-db/commit/1ac578fbb1242b7cdb619f2ad742dd3b7772ea0a#diff-ca8362b8be8f02af0316d9c0e3979808) was not right implemented as the `$pkck` variable is used before it's declared.

- [:question:] Are you fixing a bug?
  - [ ] Detail how the bug is invoked currently.
  - [ ] Detail the original, incorrect behavior.
  - [ ] Detail the new, expected behavior.
  - [ ] Base your feature on the `master` branch, and submit against that branch.
  - [ ] Add a regression test that demonstrates the bug, and proves the fix.
  - [ ] Add a `CHANGELOG.md` entry for the fix.

- [ ] Are you creating a new feature?
  - [ ] Why is the new feature needed? What purpose does it serve?
  - [ ] How will users use the new feature?
  - [ ] Base your feature on the `develop` branch, and submit against that branch.
  - [ ] Add only one feature per pull request; split multiple features over multiple pull requests
  - [ ] Add tests for the new feature.
  - [ ] Add documentation for the new feature.
  - [ ] Add a `CHANGELOG.md` entry for the new feature.

- [ ] Is this related to quality assurance?
  

- [ ] Is this related to documentation?

PR #341, opened 2018-10-18
Fix undefined variable bug in MetadataFeature

Xerkus:
This PR fixes bug introduced in #315 
It was not released and exists only in master/develop branches as such no changelog entry is needed.
This PR restores original behavior, adds tests and reapplies changes from #315 properly.

- [x] Are you fixing a bug?
    Bug is invoked when metadata information is recorded into shared data of the TDG metadata feature
    Undefined variable error is thrown and invalid data is written into shared data
    Proper behavior restored, column name or list of columns in primary key are written into shared data

Supersedes #340

### DIFF
--- a/src/TableGateway/Feature/MetadataFeature.php
+++ b/src/TableGateway/Feature/MetadataFeature.php
@@ -73,11 +73,11 @@ class MetadataFeature extends AbstractFeature
             throw new Exception\RuntimeException('A primary key for this column could not be found in the metadata.');
         }
 
-        $pkck = $pkck->getColumns();
-        if (count($pkc) === 1) {
-            $primaryKey = $pkck[0];
+        $pkcColumns = $pkc->getColumns();
+        if (count($pkcColumns) === 1) {
+            $primaryKey = $pkcColumns[0];
         } else {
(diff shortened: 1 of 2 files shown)


src/TableGateway/Feature/MetadataFeature.php
```diff
@@ -73,11 +73,11 @@ public function postInitialize()
...
 
-        $pkck = $pkck->getColumns();
-        if (count($pkc) === 1) {
-            $primaryKey = $pkck[0];
+        $pkcColumns = $pkc->getColumns();
+        if (count($pkcColumns) === 1) {
+            $primaryKey = $pkcColumns[0];
         } else {
-            $primaryKey = $pkc;
+            $primaryKey = $pkcColumns;
```
Xerkus: Original line had `$pkc->getColumns()` here: https://github.com/zendframework/zend-db/commit/1ac578fbb1242b7cdb619f2ad742dd3b7772ea0a#diff-ca8362b8be8f02af0316d9c0e3979808L80

ezimuel:
@Xerkus thanks!

PR #342, opened 2018-11-12
Delete from aliased table

OleksandrBonar:
This fix possible delete items from aliased table through TableGateway class

### DIFF
--- a/src/TableGateway/AbstractTableGateway.php
+++ b/src/TableGateway/AbstractTableGateway.php
@@ -439,10 +439,7 @@ abstract class AbstractTableGateway implements TableGatewayInterface
     protected function executeDelete(Delete $delete)
     {
         $deleteState = $delete->getRawState();
-        if ($deleteState['table'] != $this->table
-            && (is_array($deleteState['table'])
-                && end($deleteState['table']) != $this->table)
-        ) {
+        if ($deleteState['table'] != $this->table) {
             throw new Exception\RuntimeException(
                 'The table name of the provided Delete object must match that of the table'
             );
--- a/src/TableGateway/AbstractTableGateway.php
+++ b/src/TableGateway/AbstractTableGateway.php
@@ -441,7 +441,7 @@ abstract class AbstractTableGateway implements TableGatewayInterface
         $deleteState = $delete->getRawState();
         if ($deleteState['table'] != $this->table
             && (is_array($deleteState['table'])
-                && end($deleteState['table']) != $this->table
+                && end($deleteState['table']) != $this->table)
         ) {
             throw new Exception\RuntimeException(
                 'The table name of the provided Delete object must match that of the table'
--- a/src/TableGateway/AbstractTableGateway.php
+++ b/src/TableGateway/AbstractTableGateway.php
@@ -455,7 +455,7 @@ abstract class AbstractTableGateway implements TableGatewayInterface
         if (is_array($deleteState['table'])) {
             $tableData      = array_values($deleteState['table']);
             $unaliasedTable = array_shift($tableData);
-            $delete->table($unaliasedTable);
+            $delete->from($unaliasedTable);
         }
 
         $statement = $this->sql->prepareStatementForSqlObject($delete);
@@ -466,7 +466,7 @@ abstract class AbstractTableGateway implements TableGatewayInterface
 
         // Reset original table information in Delete instance, if necessary
         if ($unaliasedTable) {
-            $delete->into($deleteState['table']);
+            $delete->from($deleteState['table']);
         }
 
         return $result->getAffectedRows();
--- a/src/TableGateway/AbstractTableGateway.php
+++ b/src/TableGateway/AbstractTableGateway.php
@@ -448,6 +448,13 @@ abstract class AbstractTableGateway implements TableGatewayInterface
         // pre delete update
         $this->featureSet->apply(EventFeatureEventsInterface::EVENT_PRE_DELETE, [$delete]);
 
+        $unaliasedTable = false;
+        if (is_array($updateState['table'])) {
+            $tableData      = array_values($deleteState['table']);
+            $unaliasedTable = array_shift($tableData);
+            $delete->table($unaliasedTable);
+        }
+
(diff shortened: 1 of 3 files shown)


OleksandrBonar:
This fix possible delete items from aliased table through TableGateway class

src/TableGateway/AbstractTableGateway.php
```diff
@@ -439,7 +439,10 @@ public function deleteWith(Delete $delete)
     protected function executeDelete(Delete $delete)
     {
         $deleteState = $delete->getRawState();
-        if ($deleteState['table'] != $this->table) {
+        if ($deleteState['table'] != $this->table
+            && (is_array($deleteState['table'])
+                && end($deleteState['table']) != $this->table)
```
malukenho: May `!in_array()` applies here?

OleksandrBonar: No. The condition is true.

src/TableGateway/AbstractTableGateway.php
```diff
@@ -439,7 +439,10 @@ public function deleteWith(Delete $delete)
     protected function executeDelete(Delete $delete)
     {
         $deleteState = $delete->getRawState();
-        if ($deleteState['table'] != $this->table) {
+        if ($deleteState['table'] != $this->table
+            && (is_array($deleteState['table'])
```
webimpress: I think you want to have here `||` instead of `&&`. This should be covered in tests. I can't see related test to this change.

test/unit/TableGateway/AbstractTableGatewayTest.php
```diff
@@ -360,6 +360,48 @@ public function testDelete()
...
+        $delete1->expects($this->once())
+            ->method('getRawState')
+            ->will($this->returnValue([
+                'table' => 'foo',               // Standard table name format, valid according to Delete::from()
+                'columns' => null,
+            ]));
+        $affectedRows = $this->table->deleteWith($delete1);
+        self::assertEquals(5, $affectedRows);
+
+        // Case 2
```
webimpress: Please split test into two test cases.

OleksandrBonar: Ok

src/TableGateway/AbstractTableGateway.php
OleksandrBonar: In "executeSelect" method the same logic

webimpress: Hm... So I guess in both places it's wrong. What do you think?

test/unit/TableGateway/AbstractTableGatewayTest.php
OleksandrBonar: @webimpress Done. First test has been unnecessary.

src/TableGateway/AbstractTableGateway.php
OleksandrBonar: executeSelect for aliased table it works!

OleksandrBonar:
@malukenho, @webimpress what next, guys?

src/TableGateway/AbstractTableGateway.php
webimpress: @OleksandrBonar You are missing test to cover that change.

webimpress:
@OleksandrBonar 

I'd really like to help, but I am not the maintainer of the library... You have to be patient :)
Please add more test to cover all changes you made. See my comment above.

src/TableGateway/AbstractTableGateway.php
ezimuel: @OleksandrBonar can you provide the unit test mentioned by @webimpress? Thanks!

OleksandrBonar:
@webimpress, @ezimuel Tests have already been added. What need to add?

OleksandrBonar:
@webimpress Method **testDeleteWithArrayTable** in file **test/unit/TableGateway/AbstractTableGatewayTest.php** must catch an exception

webimpress:
@OleksandrBonar
> Method `testDeleteWithArrayTable` in file `test/unit/TableGateway/AbstractTableGatewayTest.php` must catch an exception

 As I can see that method tests just positive scenario, without the exception. I'd like to have scenarios when as the result we should get the exception so something with:

```php
$this->expectException(...);
$this->expectExceptionMessage(...);
```

and it is not there, but it should be.

OleksandrBonar:
I apologize. Looking at a fresh look at the code, I realized that checking the table variable was overly complicated.

OleksandrBonar:
@webimpress In the delete method, when the table name is an array with an alias, the original name is retrieved to fulfill the query. Then the alias is restored. This situation is tested in the file TableGatewayTest.php

OleksandrBonar:
@webimpress done! )

ThomasRo:
The provided solution is consistent with the one for the update statement (method AbstractTableGateway::executeUpdate), but i do not like both of them. 
They both disable the possibility to use aliases with update and delete statements, while indeed aliases make sense, in particular if you make use of additional joins with your update or delete statements. zend-db natively does allow joins in update statements; some RDBMS do also allow joins in delete statements and the zend sources can easily extended to make use of them.
I would suggest to fix Zend\Db\Sql\AbstractSql::resolveTable in order to support aliases in all statements. 
Overwriting attributes (and losing information) while executing a statement, does not feel right. Furthermore a statements gets build, which is different from that one, which you get by calling AbstractSql::getSqlString (for debugging purposes) because there, the overwriting does not take place.

webimpress:
@ThomasRo Can you provide test case, please? Then we will try to find solution to resolve both issues.

ThomasRo:
@webimpress I have tried to, but came across many impacts to other statements so that the necessary efforts are to big... Since i won't deliver a better solution, i go with the one of @OleksandrBonar

OleksandrBonar:
@ezimuel updates published

OleksandrBonar:
bump

OleksandrBonar:
@malukenho please review and approve or decline these changes

michalbundyra:
Thanks, @OleksandrBonar!

PR #349, opened 2019-01-20
Add Expression as allowed value to Select

andreasschroth:
Static analysis fails if an Expression object is passed to some methods, thus the updated docblock.


ezimuel:
Thanks @andreasschroth

PR #356, opened 2019-02-12
add support for `IN (NULL)` for empty value-set

pine3ree:
Calling `$expression->in($identifier, [])` results in an invalid SQL string.
For an empty valueSet we should return a never-satisfied `(IVL)` condition string such as `(NULL)`.


michalbundyra:
Thanks, @pine3ree!

PR #358, opened 2019-02-18
(just) rename a variable for clarity

pine3ree:
Use singular name when creating a new predicate expression for `addPredicate($predicate)`, especially inside the loop. Leave it alone when the `$predicates` argument is used directly.


michalbundyra:
Thanks, @pine3ree!

PR #359, opened 2019-02-19
update pdo-result current() doc-block

pine3ree:
`PDOStatement::fetch(...)` returns mixed type(s), `false` on failure (The doc-block for the returned property is correctly set to `mixed`)


michalbundyra:
Thanks, @pine3ree!

PR #360, opened 2019-02-19
add missing magic property doc-block entry

pine3ree:
no comment needed here, self-explanatory


michalbundyra:
Thanks, @pine3ree!

PR #384, opened 2019-07-10
Fixes SQL expression unit test

albertor24:
Fixes a typo in the title of a test, and also performs an assertion instead of just calling the function.


michalbundyra:
Thanks, @albertor24!

PR #389, opened 2019-07-29
Update `join()` documentation: First parameter can be instance of `TableIdentifier`

jbelien:
Update `join()` documentation: First parameter can be instance of `TableIdentifier`.
See https://github.com/zendframework/zend-db/blob/master/src/Sql/Select.php#L265 (**updated**)

- [x] Is this related to documentation?


michalbundyra:
Thanks, @jbelien!

Issue #385, opened 2019-07-10
\Zend\Db\Adapter\Driver\Pdo\Connection::rollback error


sorkl:
https://github.com/zendframework/zend-db/blob/4c68f2c33beb76d63a59c5bd5e6c96c9763966d7/src/Adapter/Driver/Pdo/Connection.php#L345
the function  code like bellow  maybe  more better
```
    public function rollback()
    {
        if (!$this->isConnected()) {
            throw new Exception\RuntimeException('Must be connected before you can rollback');
        }

        if (!$this->inTransaction()) {
            throw new Exception\RuntimeException('Must call beginTransaction() before you can rollback');
        }

        if (0 === $this->nestedTransactionsCount) {
            $this->resource->rollBack();
            $this->nestedTransactionsCount = 0;
        }

        return $this;
    }
```

webimpress:
@sorkl Would you like submit PR with the change and unit test? Thanks!

usuyukiso:
I have similar issue  
Must call beginTransaction() before you can rollback In Zend\Db\Adapter\Driver\Pdo\Connection.php line 353.
Even though it was transaction. Also checked that it works with php 5.6, but fails with php7.2

I use libraries:
zendframework/zend-db     2.10.0
zendframework/zend-stdlib 3.2.1

michalbundyra:
See: #390. Closing as "won't fix".

PR #390, opened 2019-07-29
Fix nested transaction issue with PDO driver

### LINKED ISSUE #197: Why "rollback" does not use $nestedTransactionsCount to decide when to real rollback like "commit" in Zend\Db\Adapter\Driver\Pdo\Connection?
When I use nested transactions, function "commit" will be real commit when $nestedTransactionsCount = 0 which is the last commit. But function "rollback" will be real rollback at the first rollback. Why these two functions were designed in two different ways?Thanks.

```
class Foo
{
    public function execute()
    {
        $this->dbConnection->beginTransaction();
        try {
            // do something
            // ...
            
            // call Bar::execute()
            $bar = new Bar();
            if (!$bar->execute()) {
                throw new Exception('Bar execute fail');
            }
            $this->dbConnection->commit();
        } catch (Exception $ex) {
            $this->dbConnection->rollback();
            return false;
        }
        return true;
    }
}

class Bar
{
    public function execute()
    {
        $this->dbConnection->beginTransaction();
        try {
            // do something
            // ...
            
            // call Bar::execute()
            $bar = new Bar();
            if (!$this->prepare()) {
                throw new Exception('Bar prepare fail');
            }
            $this->dbConnection->commit();
        } catch (Exception $ex) {
            $this->dbConnection->rollback();
            return false;
        }
        return true;
    }
    
    public function prepare()
    {
        return false;
    }
}

// This will get a "Must call beginTransaction() before you can rollback" exception
// Because $inTransaction = false
$foo = new Foo();
if (!$foo->execute()) {
    echo 'fail';
} else {
    echo 'success';
}
```

michalbundyra:
Fixes #197, #385

For details please see related tickets.

/cc @danielhuang-030 @metalinspired @sorki @usuyukiso

### DIFF
--- a/src/Adapter/Driver/Pdo/Connection.php
+++ b/src/Adapter/Driver/Pdo/Connection.php
@@ -352,10 +352,11 @@ class Connection extends AbstractConnection
             throw new Exception\RuntimeException('Must call beginTransaction() before you can rollback');
         }
 
-        $this->resource->rollBack();
-
-        $this->inTransaction           = false;
-        $this->nestedTransactionsCount = 0;
+        --$this->nestedTransactionsCount;
+        if ($this->nestedTransactionsCount === 0) {
+            $this->resource->rollBack();
+            $this->inTransaction = false;
+        }
 
         return $this;
     }
--- a/src/Adapter/Driver/Pdo/Connection.php
+++ b/src/Adapter/Driver/Pdo/Connection.php
@@ -30,6 +30,11 @@ class Connection extends AbstractConnection
     protected $dsn = null;
 
     /**
+     * @var int
+     */
+    protected $internalNestedTransactionCount = 0;
+
+    /**
      * Constructor
      *
      * @param  array|\PDO|null                    $connectionParameters
@@ -307,7 +312,8 @@ class Connection extends AbstractConnection
             $this->inTransaction = true;
         }
 
-        $this->nestedTransactionsCount ++;
+        ++$this->nestedTransactionsCount;
+        ++$this->internalNestedTransactionCount;
 
         return $this;
     }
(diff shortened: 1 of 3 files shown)


webimpress:
Actually I think it is still wrong, because for case:
```
beginTransaction
  execute QUERY1
  beginTransaction
    execute QUERY2
  rollback
  execute QUERY3
commit
```

all queries are going to be committed ... Need to think about better solution for this.

webimpress:
I've added more test cases and I am even more confused and not sure if the fix is needed. Let me explain:
```
beginTransaction
  execute QUERY1
  beginTransaction
    execute QUERY2
  rollback
  execute QUERY3
commit
```
this is gonna throw an exception from the final commit as transaction is gonna be closed on the rollback. But in theory it should pass without errors, only QUERY2 should be rollbacked. But it's not possible in world without nested transactions.
- QUERY1 is rollbacked (wrong!)
- QUERY2 is rollbacked (correct)
- QUERY3 is committed (correct)

Let consider now another example:
```
beginTransaction
  execute QUERY1
  beginTransaction
    execute QUERY2
  rollback
  execute QUERY3
rollback
```

and, previously (before fix), the second rollback here was failing for the same reason as commit in the first example. Now, this rollback is not failing but:
- QUERY1 is rollbacked (correct)
- QUERY2 is rollbacked (correct)
- QUERY3 is committed (wrong!)

Based on that, I am not sure if any fix is needed. I would consider it as a limitation for PDO Driver. 

The only problem I see is that we silently allowing open nested transaction, what is not allowed in PDO. We would get the exception straight away there if we try to open another transaction if the previous is still open. If we change the behaviour then probably all will be right. I mean, it will be not possible to do:
```
beginTransaction
   execute QUERY1
   beginTransaction <-- it will fail here as transaction cannot be nested.
```

/cc @Ocramius

michalbundyra:
Closing as "won't fix", as PDO does not support nested transactions.

PR #393, opened 2019-08-05
update return type test logic

pine3ree:
- use internal property `$allowedReturnTypes` (maybe making it a static property or a class constant would be advisable)
- use strict `in_array` test as for instance this expression `in_array(0, ['a', 'b'])` evaluates to true


michalbundyra:
Thanks, @pine3ree!

PR #399, opened 2019-12-08
Hotfix #376: Accessing constraint metadata within the Oracle adapter

### LINKED ISSUE #376: Oracle stuff
I wonder if I did something wrong.

I cannot get constraints with OracleMetadata. Maybe the following code is missing in loadConstraintData method.

$this->data['constraints'][$schema][$table] = $constraints;

And the method \Zend\Db\Adapter\Platform\Oracle::setDriver.
The following code may be "$driver instanceof \PDO":

$driver instanceof PDO && $driver->getAttribute(PDO::ATTR_DRIVER_NAME) == 'oci')

panychek:
Fixes #376 

The `OracleMetadata::loadConstraintData` method simply does not work:
```php
$metadata = \Zend\Db\Metadata\Source\Factory::createSourceFromAdapter($adapter);
$constraints = $metadata->getConstraints($table, $schema); // always an empty array
```

The existing test couldn't catch it because there is no real table involved:
https://github.com/zendframework/zend-db/blob/4c68f2c33beb76d63a59c5bd5e6c96c9763966d7/test/unit/Metadata/Source/OracleMetadataTest.php#L48-L49

### DIFF
--- a/src/Metadata/Source/OracleMetadata.php
+++ b/src/Metadata/Source/OracleMetadata.php
@@ -172,6 +172,8 @@ class OracleMetadata extends AbstractSource
             }
         }
 
+        $this->data['constraints'][$schema][$table] = $constraints;
+
         return $this;
     }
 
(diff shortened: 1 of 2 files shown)


michalbundyra:
@panychek Thanks for your contribution. We need unit test for that change.

panychek:
I modified the existing one, so please have a look.
BTW, there are no tests for this behavior in either of the adapters.

michalbundyra:
Thanks, @panychek!

